PR #5, opened 2016-09-23
Fixed license headers, replaced GPL license with BSD

jperedadnr:
Added third par…ties license


Birdasaur:
I agree to this merge but the web merge option is not working. Will merge soon or Jose can merge.

PR #10, opened 2016-11-12
Update .gitignore

miho:
Temporary files in `.gradle` folder should not be committed. Ony `gradle` folder should be committed if the gradle wrapper is used.


Birdasaur:
Merged. Please verify.

PR #12, opened 2016-11-14
PolyLine3D and ControlsFX dependency upgrade

Birdasaur:
PolyLine3D class now supports both Ribbon and Triangle Tube mesh pattern.
Legacy Constructor to maintain backwards compatibility uses Ribbon mesh.
PolyLines3D sample updated to use new constructor.
ControlsFX dependency upgraded to 8.40.12 in gradle script

### DIFF
--- a/FXyz-Core/src/main/java/org/fxyz/shapes/composites/PolyLine3D.java
+++ b/FXyz-Core/src/main/java/org/fxyz/shapes/composites/PolyLine3D.java
@@ -57,27 +57,7 @@ public class PolyLine3D extends Group {
     
     //Creates a Ribbon PolyLine3D
     public PolyLine3D(List<Point3D> points, int width, Color color) {
-        this.points = points;
-        this.width = width;
-        this.color = color;
-        setDepthTest(DepthTest.ENABLE);        
-        mesh  = new TriangleMesh();
-        buildRibbon();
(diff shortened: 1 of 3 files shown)


FXyz-Core/src/main/java/org/fxyz/shapes/composites/PolyLine3D.java
```diff
@@ -27,72 +27,137 @@
...
+    public List<Point3D> points;
+    public int width = 2;
+    public Color color = Color.WHITE;
+    private TriangleMesh mesh;
+    public MeshView meshView;
+    public PhongMaterial material;
+    public static enum LineType {RIBBON, TRIANGLE};
+    
+    //Creates a Ribbon PolyLine3D
+    public PolyLine3D(List<Point3D> points, int width, Color color) {
```
Birdasaur: Oh crap yes you are right. I forgot to fix that. Will fix and issue new pull request

jperedadnr: No need for new PR, just fix it on your fork

Birdasaur: Will fix, commit and let you know when ready to merge.

PR #17, opened 2017-10-18
Migration of SphereSegment to SegmentedDomeMesh with Sample

Birdasaur:
Removed old SphereSegment class.
Created new SegmentedDomeMesh class.
Added SegmentedDome sample which uses standard TexturedMesh controls to manipulate the geometry.
Added RandomSegmentedDomes sample which was really fun and I didn't want to just throw away.


FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/SegmentedDomeMesh.java
```diff
@@ -0,0 +1,231 @@
...
+                        (float) (radius * Math.cos(theta) * Math.cos(phi)),
+                        (float) (radius * Math.sin(theta)));
+                mesh.getPoints().addAll(new Float(p3D.getX()), new Float(p3D.getY()), new Float(p3D.getZ()));
+                theta += (thetamax - thetamin) / granularity;
+            }
+            phi += (phimax - phimin) / granularity;
+        }
+
+        //for now we'll just make an empty texCoordinate group
+        mesh.getTexCoords().addAll(0, 0);
```
jperedadnr: Any plan to add the texture coordinates soon? It doesn't really help having the primitive extending TexturedMesh otherwise. I'd suggest having a look at how SegmentedSphereMesh does it.

jperedadnr: Important: regardless the value of the texture coordinates, having a TexturedMesh we need to use its objects like `ListVertices`, or `textureCoords`, otherwise when rendering the sample we'll have NPE or empty results. 
Again, you can check how the SegmentedSphere sets those lists.

Birdasaur: Texture Coordinates have been added.

Birdasaur: This has been implemented.

Birdasaur:
Changes requested in review have been implemented.

PR #39, opened 2018-10-06
issue-36: Create branch 11.0.0

jperedadnr:
- Source/Target compatibility bumped to 1.9
- Core can be used as dependency on any project with Java 9, 10, 11
- Sampler runs standalone in JDK: 9, 10, 11
- Added Launcher class to Sampler to allow running via shadow jar
- Remove private API

### DIFF
--- a/FXyz-Samples/build.gradle
+++ b/FXyz-Samples/build.gradle
@@ -34,28 +34,31 @@ mainClassName = fxyz_sampler_mainClass
 ext.mainClass = mainClassName
 applicationDefaultJvmArgs = ["-XX:+UseConcMarkSweepGC"]
 
-configurations {
-    jdk
-}
-
-sourceSets {
-    main {
-        compileClasspath += configurations.jdk
-    }
-}
-
 dependencies {
+    compile "org.openjfx:javafx-graphics:11:win"
+    compile "org.openjfx:javafx-graphics:11:linux"
+    compile "org.openjfx:javafx-graphics:11:mac"
+        
     compile project(':FXyz-Client')
     compile project(':FXyz-Core')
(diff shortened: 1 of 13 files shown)


Birdasaur:
Excellent work as usual Jose.

FXyz-Samples/build.gradle
```diff
@@ -34,36 +38,50 @@ mainClassName = fxyz_sampler_mainClass
...
-sourceSets {
-    main {
-        compileClasspath += configurations.jdk
-    }
-}
-
 dependencies {
+    compile "org.openjfx:javafx-graphics:11:win"
+    compile "org.openjfx:javafx-graphics:11:linux"
+    compile "org.openjfx:javafx-graphics:11:mac"
```
abhinayagarwal: Is adding all 3 platform a requirement for the fat jar?

build.gradle
```diff
@@ -25,14 +35,17 @@ subprojects {
...
 
-    sourceCompatibility = '1.8'
-    targetCompatibility = '1.8'
-
     repositories {
         jcenter()
     }
     
     dependencies {     
+        compile "org.openjfx:javafx-base:11:${platform}"
```
abhinayagarwal: You can get rid of the curly braces surrounding platform

FXyz-Samples/build.gradle
jperedadnr: Yes, this adds the native libs of all three platforms to the fat jar.
If you don't do this, when you create the jar you will have only the current platform, so the jar won't be cross-platform

abhinayagarwal: Thanks for clarifying.

PR #44, opened 2018-10-28
Issue 38: Added Maya Importer

jperedadnr:


nlisker:
There's a problem with the client where whenever you click on "Importmaya" on the sidebar it creates a new model and they stack on top of each other: 

image

jperedadnr:
Yes, thanks for reporting. The issue is related to the animation. I need to kill it when another sample is selected.

jperedadnr:
Actually the issue has nothing to do with the animation. There is a memory leak and the old model is not removed properly.

FXyz-Core/build.gradle
```diff
@@ -31,6 +31,5 @@ compileJava {
 javadoc {
     options.addBooleanOption("Xdoclint:none").setValue(true);
     options.addBooleanOption("javafx").setValue(true);
-    options.addBooleanOption("html5").setValue(true);
 }
```
nlisker: Extra spaces on the last line.

FXyz-Importers/build.gradle
```diff
@@ -35,6 +35,5 @@ compileJava {
 javadoc {
     options.addBooleanOption("Xdoclint:none").setValue(true);
     options.addBooleanOption("javafx").setValue(true);
-    options.addBooleanOption("html5").setValue(true);
 }
```
nlisker: Extra spaces on the last line.

jperedadnr:
I tend to use the invalidation listener when I want to remove it after certain condition is met. It would work the same with the change listener of course, but it is more verbose.
I'll use the same to remove the timeline.

PR #45, opened 2018-10-29
Warnings cleanup

nlisker:
A few trivial warning fixes in an effort to keep a clean codebase.

I left unused methods and missing serialVersionUID warnings alone for now.

### DIFF
--- a/FXyz-Importers/src/main/java/org/fxyz3d/importers/ImporterFinder.java
+++ b/FXyz-Importers/src/main/java/org/fxyz3d/importers/ImporterFinder.java
@@ -54,17 +54,17 @@ public class ImporterFinder {
             throw new RuntimeException("Could not import library. Failed to determine library location. URL = " + url.getPath());
         }
         File[] files = libDir.listFiles();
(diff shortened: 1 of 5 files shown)


jperedadnr:
Changes look good. 

However there are still warnings because unchecked method invocation at `Optimizer` (and deprecation at `Importer3D`, but it will be fixed with PR #44).

Not sure if there is an easy way of getting rid of those warnings at:

```
newKeyValues.add(new KeyValue((WritableValue)keyValue.getTarget(), keyValue.getEndValue(), Interpolator.DISCRETE));
```
since we don't know in advance the type of the target or end value.

nlisker:
Yeah, I left out the deprecated instantiation method because I saw you changed it in #44 and didn't want a conflict.

The `KeyValue` constructor warnings are indeed troublesome to take care of. For some reason the class was not made generic, so when its data is retrieved the type information is lost, and then we get a warning when creating a new one from that info. Some refactoring might help, but from my experience the `Optimizer` is not always functioning properly, so we'll get to that at some point.

Issue #59, opened 2019-01-20
Remove Nashorn Engine


jperedadnr:
It is deprecated and it will be removed:
> Warning: Nashorn engine is planned to be removed from a future JDK release

We need to find a replacement to evaluate expressions.

PR #60, opened 2019-01-20
Modularize FXyz

jperedadnr:
This PR will convert all FXyz sub projects into modules.

As a first consequence, source/target level has to be set to 1.9, and Core/Importers modules won't run on Java 8

Using modules, the classloader system used to scan the importers or the samples has been modified, and it doesn't make sense use a shadow plugin anymore, so the `jlink` plugin has been added.

In a follow-up PR a pipeline for distributing the FXyz Sampler can be implemented for every platform, either with `jlink` or with `jpackage`.


Birdasaur:
I went through the diff of all the changed files and I see no problems, at least in terms of becoming a modular project.  I do not have much experience with Travis or CI systems in general so I cannot comment on that.  So other than the Travis build error thing, I would approve this pull request.

nlisker:
Sorry for the delay. I'm getting an issue with this patch when refreshing gradle:

```
> Configure project :
Found module name 'org.fxyz3d.client'
Found module name 'org.fxyz3d.core'
Found module name 'org.fxyz3d.importers'
Found module name 'org.fxyz3d.FXyz.Samples'

FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\Nir\git\FXyz\FXyz-Client\build.gradle' line: 2

* What went wrong:
A problem occurred evaluating script.
> java.lang.StackOverflowError (no error message)
```

Line 2 on FXyz-Client\build.gradle gives the error:

```
Could not run build action using Gradle distribution 'https://services.gradle.org/distributions/gradle-5.0-all.zip'.
Build file 'C:\Users\Nir\git\FXyz\FXyz-Client\build.gradle' line: 2
A problem occurred evaluating script.
```

Not sure if this is a local issue. I'm on Eclipse 4.9 with Buildship 3.0.

jperedadnr:
I can open the project from Eclipse 2018-12 (Buildship 3), having `org.gradle.java.home` set to JDK 11 in my user gradle.properties file.

There were some issues building it though, so I've moved the deprecated `compile`  to `implementation`, and it all works fine now from Eclipse and from a terminal.

nlisker:
Thanks, error is now gone.

Now the issue is that Gradle puts the "Project and External Dependencies" on the Classpath instead of the Modulepath, which causes errors with finding modules. For example, in the module-info of org.fxyz3d.core:

```
requires jcsg;
requires vvecmath;
requires poly2tri.core;
```

all give the error `X cannot be resolved to a module`. This is apparently a known Eclipse issue with workarounds:
https://github.com/eclipse/buildship/issues/620
https://github.com/eclipse/buildship/issues/708
https://github.com/eclipse/buildship/issues/658

jperedadnr:
The [HelloFX modular sample for Eclipse](https://github.com/openjfx/samples/blob/master/IDE/Eclipse/Modular/Gradle/hellofx/build.gradle#L16) already uses that fix you have linked. I've just add it to the project.

nlisker:
I'm getting an error for `org.fxyz3d.samples.utilities.RayShooting`: can't access the import `com.sun.javafx.scene.CameraHelper`.
If it works for other IDEs then it means Eclipse will require its own configuration of .classpath file where `javafx.graphics` exports `com.sun.javafx.scene`.

image


Are we keeping these config files in the repo?

jperedadnr:
I'd rather not add IDE specific files to the repository. It is a pure gradle project, that should work on any IDE.
The samples build already includes the options to open the private API (compileJava and Run tasks), that should be enough to build and run using gradle tasks, shouldn't it?

(Anyway I was thinking on removing the RayTracing sample as long we shouldn't use private API)

build.gradle
```diff
@@ -72,6 +76,24 @@ subprojects {
...
+            fxml   = 'XML_STYLE'
+        }
+
+        ext.title           = 'F(X)yz'
+        ext.yearCurrent     = new SimpleDateFormat("yyyy").format(new Date())
+        ext.yearSince       = '2013'
+
+        excludes(['**/*.txt', '**/*.html', '**/*.fxml', '**/*.css', '**/*.obj', '**/*.mtl', '**/*.ma', '**/*.png', '**/*.gif', '**/*.jpg'])
+    }
 }
```
nlisker: Empty lines at end of file.

nlisker:
> I'd rather not add IDE specific files to the repository. It is a pure gradle project, that should work on any IDE.

That's fine.

> The samples build already includes the options to open the private API (compileJava and Run tasks), that should be enough to build and run using gradle tasks, shouldn't it?

I see,
```
        jvmArgs = ['--add-exports', 'javafx.base/com.sun.javafx.runtime=controlsfx', 
                   '--add-exports', 'javafx.graphics/com.sun.javafx.scene=org.fxyz3d.FXyz.Samples']
```
It should be enough to run with gradle, yes. But if someone develops in Eclipse they will have a build error until they do a manual configuration. We can add that in the readme.

> (Anyway I was thinking on removing the RayTracing sample as long we shouldn't use private API)

I haven't looked at that samples. The `CameraHelper` class actually holds method that would be useful as public API. See https://bugs.openjdk.java.net/browse/JDK-8091210. I have a plan to work on that one.

---

The first issue I described with `Could not run build action using Gradle distribution 'https://services.gradle.org/distributions/gradle-5.0-all.zip'.` came back. The cause is the line `apply from: 'gradle/publishing.gradle'` in the build.grade files. I don't know why. I think that the change from `compile` to `implementation` didn't actually do anything.

I couldn't do testings because of this, but looking at the changes, the patch looks good, so I think you can commit.

jperedadnr:
It would be great to have API for the CameraHelper methods. In fact, the last comment on the issue you have linked was probably done in regard to the RayShooting sample (same author).

compile/implementation should have nothing to do with the gradle/publishing include, so maybe you can comment that line and check again?

nlisker:
> compile/implementation should have nothing to do with the gradle/publishing include, so maybe you can comment that line and check again?

Every project for which I remove that line builds fine. I thought that the compile/implementation helped because the problem disappeared after that change, but it came back. It's probably a local machine issue or a problem with Buildship and Eclipse.

Iv'e also found that there is a seemingly unused `requires jdk.scripting.nashorn;` for `module org.fxyz3d.FXyz.Samples`. Is it there for something?

jperedadnr:
Yes, that's the Nashorn engine, it is used in the sampler to introduce math expressions. 
There is a issue #59 filed to remove it when possible (maybe with graaljs).

nlisker:
I see now, it's runtime-only.

nlisker:
After a gradle refresh `Project and External dependencies` are still put on the classpath. This causes the following errors in:

* `module org.fxyz3d.importers` for `requires org.fxyz3d.core`
* `module org.fxyz3d.FXyz.Samples` for:
```
    requires org.fxyz3d.core;
    requires org.fxyz3d.importers;
    requires org.fxyz3d.client;
    provides org.fxyz3d.FXyzSamplerProject
```

Previous errors in `module org.fxyz3d.FXyz.Samples` for:
```
    requires controlsfx;
    requires easybind;
    requires reactfx;
```
are gone though. So only fxyz modules are affected.

jperedadnr:
@nlisker The last commit should fix your issues.

Issue #38, opened 2018-10-06
New Artifact: FXyz-Importers


jperedadnr:
Create new artifact that will provide loaders/importers of complex 3D models with different formats

See https://github.com/javafxports/openjdk-jfx/issues/203

As a first step, the 3DViewer importers can be included:
- .obj
- .dae
- .maya
- .max

nlisker:
Didn't see this until now (you might want to notify on the openjfx issue). Sign me up!

.obj is by far the simplest format, supporting only mesh data and references to external materials. The other ones support animations, lighting etc. in addition.

What to import?
* Mesh data is of course a given.
* Animations will have to be imported since there's no way to contain them externally. In 3DViewer, only .ma files import animations, and only using quads (instead of the built in trigs of JavaFX). Tests Iv'e done show very poor performance of these animations.
* Materials can be imported directly, but that would require converting everything to Phong, which might cause some issues. Materials can be stored externally (via image files) and added in the javaFX side..
* Lighting will have to be converted to `PointLight`, or not imported at all. Lighting can be added in the JavaFX side.

As mentioned earlier, 3DViewer contains groundwork for polygon meshes (quads). It would be very nice to have this option as another subclass of `Mesh`, but that seems like extensive work if done properly, and maybe out of scope.

jperedadnr:
About quads, the plan is to add them to FXyz-Core, that contains primitives and other shapes based on TriangleMesh. After all, the `PolygonMesh`/`PolygonMeshView` [approach](https://github.com/javafxports/openjdk-jfx/blob/develop/apps/samples/3DViewer/src/main/java/com/javafx/experiments/shape3d/PolygonMeshView.java#L276) is just _another _TriangleMesh_.  While it is a smart implementation and good enough for now, having other `Mesh` implementations in JavaFX would be great, but as you said, out of scope probably.

I'll try to create a first PR with the obj importer soon, including changes in FXyz-Core, and a sample in FXyz-Samples. And I'll notify on the openjfx issue of course.

jperedadnr:
@nlisker I've created a [WIP] PR with [FXyz-Importers](https://github.com/FXyz/FXyz/pull/42), including obj for now. There is also a sample that can be run from FXyz-Samples.
I've added you as collaborator of the project, so I can add you as reviewer.

jperedadnr:
@nlisker The next two importers are DAE and MAX

[...]

PR #63, opened 2019-01-31
Redesign Importer API

### LINKED ISSUE #61: MtlReader NPE when ObjImporter constructed using InputStream
When `.obj` has mtl and ObjImporter was constructed through `public ObjImporter(InputStream inputStream)` then `private String objFileUrl` is never initialized, hence during mtl reading, `MtlReader()` throws NPE

AlmasB:
- Fixes #61 
- Merged ObjImporter and PolyImporter and their wrappers
- Removed duplicated code in above classes
- Added general data structure Model
- Importer is now an interface with a slightly different API

ObjImporter + MtlReader can be refactored further, but since this is a major change it would be good to get a review of the overall API earlier rather than later.


FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
```diff
@@ -0,0 +1,55 @@
+package org.fxyz3d.importers;
```
jperedadnr: The license header is missing

jperedadnr: I'd use the FXyz license in this case, updated to 2019.

AlmasB: I'll grab the LICENSE one then, with 2013-2019

jperedadnr:
There are two samples that make use of the importers, those should be updated as well.

AlmasB:
You mean their license headers?

jperedadnr:
No, you have modified the API, haven't you? The samples still use a String for loading the model, so at least that would change to an URL.

jperedadnr:
Ok, I see. You still keep the strings on `Importer3D`? I assumed you were going to change this one too to use URL.

AlmasB:
I just thought we could have a review before I start changing everything. But since we are happy (at least with the URL update), I will now update everything to use URL.

jperedadnr:
Do you think it would make sense to keep `Importer3D` with the existing String and add new methods with URL?

AlmasB:
Imo no, since the call site is likely to have URL rather than String. Even the existing samples use `class.getResource()` and then get `toExternalForm()` to just create `URL` internally. Plus, if you have `File` or `Path`, I believe it's much more convenient to obtain URL from these, rather than try and guess what String format it should be.

jperedadnr:
Ok, go ahead then

AlmasB:
Couldn't we do [this](https://github.com/FXyz/FXyz/blob/master/FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer3D.java#L125) statically just once, rather than checking for importers every time we want to load a model?

Then we could do something like:
```
val importer = importers.find { it.isSupported(extension) } ?: throw IOException( ... )
importer.load()
```

jperedadnr:
With the `Importer3D` you could inject an importer dynamically, other than the existing obj/maya importers. Not that this is going to happen very often, and even if it does, I don't see the point of providing importers if you are going to use a third party importer?

Can you review the existing conflicts?

FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer.java
```diff
@@ -32,38 +32,32 @@
 package org.fxyz3d.importers;
```
jperedadnr: Make sure to update the header of modified files:
Something like:
```
 * Copyright (c) 2019 F(X)yz
 * Copyright (c) 2014, Oracle and/or its affiliates.
 ```

File: FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java [hunk 3]
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.*;
+
+/**
+ * @author Almas Baimagambetov (almaslvl@gmail.com)
+ */
+public class Model {
+
+    private final Group root = new Group();
+    private Map<String, Material> materials = new HashMap<>();
```
jperedadnr: use `final`

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
```diff
@@ -0,0 +1,83 @@
...
+    private Map<String, Node> meshViews = new HashMap<>();
```
jperedadnr: use `final`.

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
```diff
@@ -0,0 +1,83 @@
...
+import javafx.scene.Group;
+import javafx.scene.Node;
+import javafx.scene.paint.Material;
+
+import java.util.*;
+
+/**
+ * @author Almas Baimagambetov (almaslvl@gmail.com)
+ */
+public class Model {
```
jperedadnr: A small JavaDoc will be nice. I can't speak for the rest of FXyz... but we should do it eventually.

FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer.java
```diff
@@ -32,38 +32,32 @@
...
+public interface Importer {
 
-public abstract class Importer {
     /**
      * Loads the 3D file
      *
      * @param url The url of the 3D file to load
-     * @param asPolygonMesh When true load as a PolygonMesh if the loader
-     * supports.
      * @throws IOException If issue loading file
```
jperedadnr: Add missing return tag

FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer.java
```diff
@@ -32,38 +32,32 @@
...
+
     /**
-     * Gets the 3D node that was loaded earlier through the load() call
-     * @return The loaded node
+     * Loads the 3D file as a polygonal mesh.
+     *
+     * @param url The url of the 3D file to load
+     * @throws IOException If issue loading file
      */
-    public abstract Group getRoot();
```
jperedadnr: same here

FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer3D.java
```diff
@@ -90,13 +86,15 @@ public static Node load(String fileUrl, boolean asPolygonMesh) throws IOExceptio
      * @return The loaded Node which could be a MeshView or a Group and the Timeline animation
      * @throws IOException if issue loading file
      */
-    public static Pair<Node,Timeline> loadIncludingAnimation(String fileUrl, boolean asPolygonMesh) throws IOException {
+    public static Model loadIncludingAnimation(URL fileUrl, boolean asPolygonMesh) throws IOException {
+        String extForm = fileUrl.toExternalForm();
```
jperedadnr: Check for possible null URL to avoid NPE (given that URL comes from `getResource` quite often you can get an null resource)

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportOBJ.java
```diff
@@ -62,9 +62,7 @@
     @Override
     protected void createMesh() {
         try {
-            model = (Group) Importer3D.load(
-                    ImportOBJ.class.getResource("DukeKing.obj").toExternalForm(),
-                    asPolygonMesh);
+            model = Importer3D.loadAsPoly(ImportOBJ.class.getResource("/org/fxyz3d/importers/DukeKing.obj")).getRoot();
```
jperedadnr: Why did you change the path? The obj file is not in that path. It is `/org/fxyz3d/samples/importers/DukeKing.obj`, so you can use `getResource("DukeKing.obj")`

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportMaya.java
```diff
@@ -67,12 +68,12 @@
     @Override
     protected void createMesh() {
         try {
-            Pair<Node,Timeline> content = Importer3D.loadIncludingAnimation(ImportMaya.class.getResource("King_WalkCycle.ma").toExternalForm(),
+            Model modelData = Importer3D.loadIncludingAnimation(ImportMaya.class.getResource("/org/fxyz3d/importers/King_WalkCycle.ma"),
```
jperedadnr: Same here (use `getResource("King_WalkCycle.ma")`)

FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer3D.java
```diff
@@ -141,16 +139,24 @@ public static Node load(String fileUrl, boolean asPolygonMesh) throws IOExceptio
...
+                model.addMeshView("default", new MeshView((TriangleMesh) fxmlRoot));
+
+                return model;
             }
+
             throw new IOException("Unknown object in FXML file [" + fxmlRoot.getClass().getName() + "]");
         } else {
-            importer.load(fileUrl, asPolygonMesh);
-            return new Pair<>(importer.getRoot(), importer.getTimeline());
+
```
jperedadnr: Remove blank line

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportOBJ.java
AlmasB: I don't believe I did. Not sure, maybe different old version?

jperedadnr: Samples are [here](https://github.com/FXyz/FXyz/tree/master/FXyz-Samples/src/main/resources/org/fxyz3d/samples/importers), don't you have this? Does the sampler work for you if you load these?

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2014, Oracle and/or its affiliates.
```
jperedadnr: This is a big refactoring. It looks fine, but have you tried loading different samples (obj/polyObj)?

AlmasB: I tried loading my own obj files. They load fine.

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportOBJ.java
AlmasB: There were probably changes to this after I forked. Will fix

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
```diff
@@ -0,0 +1,100 @@
...
+import javafx.scene.paint.Material;
+
+import java.util.*;
+
+/**
+ * Represents a loader-independent 3D model data.
+ *
+ * @author Almas Baimagambetov (almaslvl@gmail.com)
+ */
+public class Model {
```
jperedadnr: It would probably make sense to rename the class to `Model3D`, wouldn't it?

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private final Group root = new Group();
+    private final Map<String, Material> materials = new HashMap<>();
+    private final Map<String, Node> meshViews = new HashMap<>();
+
+    /**
+     * The root that may contain Node / MeshView / PolygonMeshView.
+     * You can add this root to your scene graph.
```
jperedadnr: Rephrase to: `This root can be added to the scene graph.`

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public final Material getMaterial(String key) {
+        return materials.get(key);
+    }
+
+    public final List<Material> getMaterials() {
+        return new ArrayList<>(materials.values());
+    }
+
+    public Optional<Timeline> getTimeline() {
```
jperedadnr: Complete JavaDoc?

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
```diff
@@ -31,25 +31,28 @@
  */
```
jperedadnr: update license header

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
```diff
@@ -75,7 +78,7 @@ public MayaGroup getRoot() {
     // MayaImporter.getTimeline() returns a JavaFX timeline
     // (javafx.animation.Timeline)
     //=========================================================================
-    @Override
+
     public Timeline getTimeline() {
```
jperedadnr: since now we have a model, I'd remove these two methods `getRoot()` and `getTimeline()`.

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
```diff
@@ -87,14 +90,46 @@ public Timeline getTimeline() {
         return meshParents;
     }
 
+    @Override
+    public Model load(URL url) throws IOException {
+        load(url, false);
+        return new Model() {
+            @Override
+            public Optional<Timeline> getTimeline() {
+                return Optional.of(MayaImporter.this.getTimeline());
```
jperedadnr: `this.timeline`

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
```diff
@@ -87,14 +90,46 @@ public Timeline getTimeline() {
...
+        };
+    }
+
+    @Override
+    public Model loadAsPoly(URL url) throws IOException {
+        load(url, true);
+        return new Model() {
+            @Override
+            public Optional<Timeline> getTimeline() {
+                return Optional.of(MayaImporter.this.getTimeline());
```
jperedadnr: `this.timeline`

AlmasB:
What's the difference between `load()` and `loadAsPoly()`? I've previously written `.obj` readers but there was no need to handle anything separately. Shouldn't we have just a single `Model3D load()`?

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
jperedadnr: Hmm, that really makes sense, so far we don't have nothing alike for any of the "shapes" we can generate, but definitely it should be there.

jperedadnr:
In theory when parsing the obj file we should be able to find out if a regular TriangleMesh can be used (3 vertices/face), or a PolygonalMesh (4+ vertices/face). But probably that happens too late, and it seems more convenient to know it in advance. Probably we can revisit this issue later on

AlmasB:
Ok. I'm happy to let it be for now. However, I do have working code that deals with this at runtime, so there is no issue there. 

Also, I don't think clients should need to know about this. If I were a client, I would want `Model3D model = Importer3D.load( ... )` and let the framework figure out the rest.

FXyz-Importers/src/main/java/org/fxyz3d/importers/Importer3D.java
```diff
@@ -90,13 +87,18 @@ public static Node load(String fileUrl, boolean asPolygonMesh) throws IOExceptio
      * @return The loaded Node which could be a MeshView or a Group and the Timeline animation
      * @throws IOException if issue loading file
      */
-    public static Pair<Node,Timeline> loadIncludingAnimation(String fileUrl, boolean asPolygonMesh) throws IOException {
+    public static Model loadIncludingAnimation(URL fileUrl, boolean asPolygonMesh) throws IOException {
```
jperedadnr: Having a `load`/`loadAsPublic method that will call `loadIncludingAnimation`, I'd make the latter private?

AlmasB: agreed

jperedadnr:
We can postpone it for later, but that will imply changing the API we are about to define now. Do you think we should add it now?

FXyz-Importers/src/main/java/org/fxyz3d/importers/Model.java
AlmasB: I'm going to keep Model3D inside this module as part of this PR. You can then move it to the core module as needed.

AlmasB:
For just `load()`:

It seems `MayaImporter` simply ignores the `asPolygonMesh` flag in `Loader` anyway. So that's easy to refactor.

As for `ObjImporter`, there is quite a bit more of refactoring needed. So far I've just moved stuff around and didn't change much in the parsing code. Shame there aren't any unit tests for imports. Should we add some tests now, and then see if I can add runtime poly handling?

jperedadnr:
I was thinking exactly the same... but we don't have tests at all. So that's for another PR I guess

AlmasB:
If we are done on this PR, I can start gathering some different .obj cases and write tests for them

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -89,328 +70,581 @@ static void log(String string) {
...
+
+        } else {
+
+            String[] split = line.substring(2).trim().split(" +");
+            int[] faceIndexes = new int[split.length*2];
+            int[] faceNormalIndexes = new int[split.length];
+            for (int i = 0; i < split.length; i++) {
+                String[] split2 = split[i].split("/");
+                faceIndexes[i*2] = model.vertexIndex(Integer.parseInt(split2[0]));
+                faceIndexes[(i*2)+1] = (split2.length > 1 && split2[1].length()>0) ? model.uvIndex(Integer.parseInt(split2[1])) : -1;
```
nlisker: Add spaces around `>`: `split2[1].length() > 0` (for the next line too).

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -89,328 +70,581 @@ static void log(String string) {
...
+            }
+
+            ((PolyObjModel) model).facesPolygon.add(faceIndexes);
+            ((PolyObjModel) model).faceNormalsPolygon.add(faceNormalIndexes);
+            model.smoothingGroups.add(model.currentSmoothGroup);
+        }
+    }
+
+    private static void parseS(String line, ObjModel model) {
+        if (line.substring(2).equals("off")) {
```
nlisker: Consider using a ternary expression like `model.currentSmoothGroup = line.substring(2).equals("off") ? 0 : Integer.parseInt(line.substring(2));`

jperedadnr:
@nlisker If you were planning to review this PR, we can wait, otherwise it's okay, we'll merge it now so we can start working on a few pending ideas: moving `Model3D` to Core, removing `asPoly`, adding tests...

nlisker:
I'm in the process of reviewing. Will finish in an hour or so.

jperedadnr:
Great, take your time!

AlmasB:
@jperedadnr I was thinking I could base the refactored ObjImporter on [this](https://github.com/AlmasB/FX3D-Importers/blob/master/src/main/kotlin/com/almasb/fx3di/obj/ObjImporter.kt), which I built using the .obj file spec. It correctly handles tri and poly cases. The main reason being, there is a lot of stuff in current ObjImporter and I don't know what most of it does and I'm concerned I might break something

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -89,328 +70,581 @@ static void log(String string) {
...
+
+    private static class ObjModel extends Model3D {
+
+        List<Map<String, Material>> materialLibrary = new ArrayList<>();
+
+        FloatArrayList vertices = new FloatArrayList();
+        FloatArrayList uvs = new FloatArrayList();
+        FloatArrayList normals = new FloatArrayList();
+        IntegerArrayList smoothingGroups = new IntegerArrayList();
+        Material material = new PhongMaterial(Color.WHITE);
```
nlisker: Constructor argument not required, but OK.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -89,328 +70,581 @@ static void log(String string) {
...
+            MeshView meshView = new MeshView();
+            meshView.setId(key);
+            meshView.setMaterial(getMaterial(key));
+            meshView.setMesh(meshes.get(key));
+            meshView.setCullFace(CullFace.NONE);
+            return meshView;
+        }
+
+        void loadComplete() {
+            meshNames.forEach(meshName -> {
```
nlisker: The `{... }` scope is not needed since it's 1 line, but OK.

nlisker:
I see that `ObjImporter` is not problematic only for me. Looks like we'll have to deal with it as a separate PR.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -89,328 +70,581 @@ static void log(String string) {
...
-    public ObjImporter(InputStream inputStream) throws IOException {
-        read(inputStream);
+    public static void setScale(float scale) {
+        ObjImporter.scale = scale;
     }
 
-    public TriangleMesh getMesh() {
-        return meshes.values().iterator().next();
-    }
+    private static Map<Predicate<String>, BiConsumer<String, ObjModel>> parsers = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
AlmasB: I believe `LinkedHashMap` keeps the order of entries, including for key sets, which is how the line conditions are checked. Without a pre-defined iteration order, `l -> true` might run first and hence skip that line.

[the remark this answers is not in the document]
nlisker: `LinkedHashMap` does keep entry order, what I missed was that there is a "default" `l -> true`, but I think that if nothing matched it should be handled after the `for`, that is, `break` wasn't reached. So line 135-136 will have the log operation. Does that make sense?

AlmasB:
Thanks for this. Replies are in line. Please see below.

> I like this. Some comments in addition to the inline ones:
> 
> 1. The `isSupported` methods can be changed from `return extension != null && extension.equals("ma");` to `return "ma".equals(extension );` to circumvent the explicit `null` check. The supported file extension should also be a constant declared on the top of the class. We should also be prepared to have multiple extensions supported by the a single importer. I know you didn't touch these, but since you are already there...

Fair enough :)

> 2. On the API side of loading as poly vs. tri mesh and with or without animations: I noted in a [comment](https://github.com/FXyz/FXyz/issues/38#issuecomment-429511612) about what can be imported. We should be prepared for all these options for the future. As I see it, a user can choose to import:
>    
>    * Mesh as poly or tri (mesh/geometric data must be imported, otherwise what's the point...)
>    * With or without materials
>    * With or without animations
>    * With or without lighting
>    * etc. (whatever file formats might support)
>    
>    Because of this, I don't think the API for with/without animations is suitable. We would either want a builder, a method full of `boolean`s, a method that accepts an array of load args (possibly `enum` constants) or something flexible of the sort.

I don't have any strong feelings pro or against this. I just thought we were moving towards the "load all we can into a high-level data structure (`Model3D`) and let the user pick what's needed". In the end, I'm happy with either.



> 3. Related to the point above, and I'm not yet sure if this is taken care of or not, we should consider that a single file we load contains more than 1 "effective" model. For example, a file can contain a model of the solar system, but in facts each planet is its own model. We should be able to allow the user to access these independently.

At the moment (for ObjImporter) a single file means a single `Model3D` object. However, each "effective" model ends up being a View of some sort, which is accessible by its name (e.g. group name) via the said `Model3D` object. So the user, at this moment, already has the ability to access these independently. Not sure about MayaImporter

> 4. There are many places with loop iterations over arrays/collections. I don't know what their sizes are, but we should consider using streams for the performance heavy ones. For example, if a model with 30k meshes is loaded, a parallel stream _might_ significantly improve performance since this process is completely or near-completely parallelizable. I'm not familiar enough with the process to recommend these hotspots.

Same.

> 
> I'll need to review `ObjImporter`, especially it's subclasses more thoroughly in the future.

I have merged ObjImporter and ObjPolyImporter into a single class since 90% of the code was the same and extracted the if/else statements. I didn't change much in the parsing code per se, but if you've got the time a review would be appreciated. Thanks.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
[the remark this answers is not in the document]
AlmasB: You mean:

```
        while ((line = br.readLine()) != null) {
            try {
                // if thread-safety is needed, keySet() can be extracted
                for (Predicate<String> condition : parsers.keySet()) {
                    if (condition.test(line)) {
                        parsers.get(condition).accept(line, model);
                        break;
                    }
                }

// log here?

            } catch (Exception e) {
                Logger.getLogger(ObjImporter.class.getName()).log(Level.SEVERE, "Failed to parse line: " + line, e);
            }
        }
```

Wouldn't that mean the log will run for all lines?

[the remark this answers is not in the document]
nlisker: Oh, right, we want to log only no-match cases. Alright, let's skip this change and revisit it when the whole line reading code is re-reviewed for parallelization.

Issue #49, opened 2018-11-04
[FXyz-Samples] Control Panel requires more controls for importers


jperedadnr:
At least a File Chooser to import 3D models from disk
Drag and drop of those 3D models
Animation controls
And other similar controls to those available on 3DViewer Settings (Y up, hierarchy, ...)

jperedadnr:
Fixed with https://github.com/FXyz/FXyz/pull/69

PR #69, opened 2019-02-05
#49 Add Animation control to sampler

jperedadnr:


FXyz-Samples/src/main/java/org/fxyz3d/controls/TimelineControl.java
```diff
@@ -0,0 +1,181 @@
...
+            if (old != null) {
+                currentTimeAsPercentage.unbind();
+                currentTxt.textProperty().unbind();
+                endTxt.textProperty().unbind();
+                bar.prefWidthProperty().unbind();
+                old.currentRateProperty().removeListener(rateListener);
+            }
+            if (t == null) {
+                timelineDisplay.setVisible(false);
+                startBtn.setDisable(true);
```
AlmasB: Instead of `setDisable(true / false)` for buttons, perhaps we can do smth like: (I don't remember the exact method call) `startBtn.disableProperty().bind(timeline.isNull())`

FXyz-Samples/src/main/java/org/fxyz3d/controls/TimelineControl.java
```diff
@@ -0,0 +1,181 @@
...
+        playBtn.setOnAction(e -> {
+            Timeline timeline = getTimeline();
+            if (playBtn.isSelected()) {
+                timeline.play();
+            } else {
+                timeline.pause();
+            }
+        });
+        ffBtn.setOnMousePressed(e -> {
+            Timeline timeline = getTimeline();
```
AlmasB: Feel free to ignore this comment: I would have added 

```
private void setTimelineRate(double rate) {
    getTimeline().setRate(rate);
}
```
and then inline all lambdas, e.g.

```
ffBtn.setOnMousePressed(e -> setTimelineRate(2));
```

jperedadnr: Makes sense

FXyz-Samples/src/main/java/org/fxyz3d/controls/TimelineControl.java
jperedadnr: There is no such binding. I'll enable/disable the HBox parent of all those buttons instead.

AlmasB: I have used [isNull()](https://docs.oracle.com/javase/8/javafx/api/javafx/beans/binding/ObjectExpression.html#isNull--)

jperedadnr:
Ups, yes, I wasn't binding the property... Fixed now

PR #72, opened 2019-02-07
ObjImporter cleanup

nlisker:
Continuing the work on ObjImporter:

* Replaced `FloatArrayList` and `IntegerArrayList` with `ObservableFloatArray` and `ObservableIntegerArray` sine they did not provide much benefit. The observable arrays are the types used by the meshes.
* Moved the parsing methods into the model class where they seem like they belong. This also allows to split `parseFace` properly.
* Reading the file is now done with a stream, which can allow parallel reading/parsing for better performance on larger files. API not included.
* Various readability changes.

There's still work here, but better to do it in smaller steps.


FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/PolygonMesh.java
```diff
@@ -41,17 +41,20 @@
...
-    private final ObservableFloatArray points = FXCollections.observableFloatArray();
-    private final ObservableFloatArray texCoords = FXCollections.observableFloatArray();
+    private final ObservableFloatArray points;
+    private final ObservableFloatArray texCoords;
     public int[][] faces = new int[0][0];
     private final ObservableIntegerArray faceSmoothingGroups = FXCollections.observableIntegerArray();
     protected int numEdgesInFaces = -1; // TODO invalidate automatically by listening to faces (whenever it is an observable)
 
-    public PolygonMesh() {}
+    public PolygonMesh() {
```
jperedadnr: Maybe use these constructors?

```
private final ObservableFloatArray points;
    private final ObservableFloatArray texCoords;
    public int[][] faces;
   ... 
    public PolygonMesh() {
        this(FXCollections.observableFloatArray(), FXCollections.observableFloatArray(), new int[0][0]);
    }

    public PolygonMesh(float[] points, float[] texCoords, int[][] faces) {
        this(FXCollections.observableFloatArray(points), FXCollections.observableFloatArray(texCoords), faces);
    }

    public PolygonMesh(ObservableFloatArray points, ObservableFloatArray texCoords, int[][] faces) {
        this.points = points;
        this.texCoords = texCoords;
        this.faces = faces;
    }
```

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -401,75 +218,67 @@ void addMesh(String key) {
...
-            FloatArrayList newUVs = new FloatArrayList(uvs.size() / 2);
-            FloatArrayList newNormals = new FloatArrayList(normals.size() / 2);
-            boolean useNormals = true;
 
             for (int i = facesStart; i < faces.size(); i += 2) {
                 int vi = faces.get(i);
-                Integer nvi = vertexMap.get(vi);
+                Integer nvi = vertexMap.putIfAbsent(vi, mesh.getPoints().size() / 3);
                 if (nvi == null) {
-                    nvi = newVertexes.size() / 3;
```
jperedadnr: Got NPE here, fix:
`nvi = mesh.getPoints().size() / 3;`

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -401,75 +218,67 @@ void addMesh(String key) {
...
-                    newVertexes.add(vertices.get(vi * 3 + 2));
+                    mesh.getPoints().addAll(vertices, vi * 3, 3);
                 }
                 faces.set(i, nvi);
 
                 int uvi = faces.get(i + 1);
-                Integer nuvi = uvMap.get(uvi);
+                Integer nuvi = uvMap.putIfAbsent(uvi, mesh.getTexCoords().size() / 2);
                 if (nuvi == null) {
-                    nuvi = newUVs.size() / 2;
```
jperedadnr: NPE, fix:
`nuvi = mesh.getTexCoords().size() / 2;`

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -401,75 +218,67 @@ void addMesh(String key) {
...
                     }
                 }
                 faces.set(i + 1, nuvi);
 
                 if (useNormals) {
                     int ni = faceNormals.get(i/2);
-                    Integer nni = normalMap.get(ni);
+                    Integer nni = normalMap.putIfAbsent(ni, newNormals.size() / 3);
                     if (nni == null) {
-                        nni = newNormals.size() / 3;
```
jperedadnr: NPE: 
`nni = newNormals.size() / 3;`

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -534,90 +466,77 @@ void addMesh(String key) {
...
+            for (int i = facesStart; i < facesPolygon.size(); i++) {
                 int[] faceIndexes = facesPolygon.get(i);
                 int[] faceNormalIndexes = faceNormalsPolygon.get(i);
-                for (int j=0;j<faceIndexes.length;j+=2){
+                for (int j = 0; j < faceIndexes.length; j += 2){
                     int vi = faceIndexes[j];
-                    Integer nvi = vertexMap.get(vi);
+                    Integer nvi = vertexMap.putIfAbsent(vi, mesh.getPoints().size() / 3);
                     if (nvi == null) {
-                        nvi = newVertexes.size() / 3;
```
jperedadnr: Fix possible NPE

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -534,90 +466,77 @@ void addMesh(String key) {
...
                     }
                     faceIndexes[j] = nvi;
-//                faces.set(i, nvi);
+//                  faces.set(i, nvi);
+
                     int uvi = faceIndexes[j+1];
-                    Integer nuvi = uvMap.get(uvi);
+                    Integer nuvi = uvMap.putIfAbsent(uvi, mesh.getTexCoords().size() / 2);
                     if (nuvi == null) {
-                        nuvi = newUVs.size() / 2;
```
jperedadnr: same here

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -534,90 +466,77 @@ void addMesh(String key) {
...
                     }
                     faceIndexes[j+1] = nuvi;
-//                faces.set(i + 1, nuvi);
+//                  faces.set(i + 1, nuvi);
 
                     int ni = faceNormalIndexes[j/2];
-                    Integer nni = normalMap.get(ni);
+                    Integer nni = normalMap.putIfAbsent(ni, newNormals.size() / 3);
                     if (nni == null) {
-                        nni = newNormals.size() / 3;
```
jperedadnr: same here

FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/PolygonMesh.java
nlisker: I tried not to touch classes outside of `ObjImporter`. e.g., I don't see a reason for `int[][] faces` to be `public`. As for the constructors, it looks like only the empty constructor is being called anyway. Do you think these other constructors will be used?

jperedadnr: It's better to have more options, imo, than to restrict them because we don't have a use case for it yet.

About faces being public, I don't like that either. Probably it can be private (or protected) with a public setter/getter?

nlisker:
There're definitely improvements we can do on that front, but for performance arrays are still the best we can get I think.

AlmasB:
RE: performance, true, but performance of what? After we send the arrays to JavaFX mesh, we longer have the control for rendering. So I guess that's primarily loading / parsing performance? Personally, I'm happy to lose whatever gains there are in loading in favour of more readable code. Would be interesting to measure the difference in performance.

jperedadnr:
For small models, there is no difference, but when trying big/huge ones, every bit counts.
Maybe we can open an issue to track this conversation (actually it would be good it we could have some kind of benchmark).

PR #74, opened 2019-02-07
Basic importer test

AlmasB:
- Added junit5 but couldn't get hamcrest to work because of --add-reads issue
- More thorough tests to follow soon when `Model3D` has a richer API


FXyz-Importers/src/test/java/org/fxyz3d/importers/obj/ObjImporterTest.java
```diff
@@ -0,0 +1,63 @@
...
+    private ObjImporter importer;
+
+    @BeforeEach
+    void setUp() {
+        importer = new ObjImporter();
+    }
+
+    @Test
+    void testExtensions() {
+        assertTrue(importer.isSupported("obj"));
```
nlisker: Now that I see this test I realize that `isSupported` should use `equalsIgnoreCase` to check the extension equality, no?

FXyz-Importers/src/test/java/org/fxyz3d/importers/obj/ObjImporterTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Test
+    void testLoad() throws Exception {
+        Model3D model = importer.load(getClass().getResource("cube.obj"));
+
+        assertEquals(1, model.getRoot().getChildren().size());
+        assertEquals(1, model.getMeshNames().size());
+        assertTrue(model.getMeshNames().contains("cube"));
+        assertSame(model.getMeshView("cube"), model.getRoot().getChildren().get(0));
+    }
```
nlisker: I think we want some check for the content of `mesh.getPoints()`, `mesh.getTexCoords()`, `mesh.getFaces()`, `mesh.getNormals()`... These are the fundamental properties of the model and we would want to make sure they are read correctly. Could have a separate method for each check.

jperedadnr: This is a good start.

Obviously we can add more tests as we go. 

For instance, we should add a material and a `mtl` file, look for a more complex obj, ...

The `TriangleMesh` has already internal checks so I don't think we should test if the model from the obj itself is a proper mesh, but if we _do_ read it correctly.

AlmasB: added mtl test and multi view obj file. Not sure about poly faces yet, as I don't know what we agreed on wrt loadAsPoly

jperedadnr: You can test an obj file with 3 or more vertices per face using loadAsPoly. For now, we let the user decide how to import the file.

AlmasB: ok

AlmasB:
Done. Would be good to add test coverage at some point

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -105,7 +105,7 @@ public Model3D loadAsPoly(URL url) throws IOException {
 
     @Override
     public boolean isSupported(String extension) {
-        return SUPPORTED_EXT.equals(extension);
+        return SUPPORTED_EXT.equalsIgnoreCase(extension);
```
jperedadnr: Since you are at it, can you add the same to the Maya importer?

AlmasB: Was planning to do that together with the Maya importer test in a separate PR

build.gradle
```diff
@@ -44,7 +44,9 @@ subprojects {
         implementation ('eu.mihosoft.vrl.jcsg:jcsg:0.5.7') { transitive = false }
         implementation 'eu.mihosoft.vvecmath:vvecmath:0.3.8'
         implementation 'org.orbisgis:poly2tri-core:0.1.2'
-        testCompile 'junit:junit:[4,)'
+        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.3.1'
+        testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.3.1'
+        testCompile 'org.junit.jupiter:junit-jupiter-params:5.3.1'
```
jperedadnr: `testImplementation` ?

FXyz-Importers/src/test/java/org/fxyz3d/importers/obj/ObjImporterTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    void testLoadAsPoly() throws Exception {
+        Model3D model = importer.load(getClass().getResource("duke_king_poly.obj"));
```
jperedadnr: `importer.loadAsPoly`?

FXyz-Importers/src/test/java/org/fxyz3d/importers/obj/ObjImporterTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    void testLoad() throws Exception {
+        Model3D model = importer.load(getClass().getResource("cube.obj"));
+
+        assertEquals(1, model.getRoot().getChildren().size());
+        assertEquals(1, model.getMeshNames().size());
+        assertTrue(model.getMeshNames().contains("cube"));
+        assertSame(model.getMeshView("cube"), model.getRoot().getChildren().get(0));
+    }
```
jperedadnr: Maybe this could be included:
```
for (Node n : model.getMeshViews()) {
            assertTrue(n instanceof MeshView);
            assertTrue(((MeshView) n).getMesh() instanceof TriangleMesh);
        }
```

FXyz-Importers/src/test/java/org/fxyz3d/importers/obj/ObjImporterTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    void testLoadAsPoly() throws Exception {
+        Model3D model = importer.load(getClass().getResource("duke_king_poly.obj"));
+
+        assertEquals(9, model.getRoot().getChildren().size());
+        assertEquals(9, model.getMeshNames().size());
+        assertEquals(9, model.getMeshViews().size());
+    }
```
jperedadnr: same here:
```
for (Node n : model.getMeshViews()) {
            assertTrue(n instanceof PolygonMeshView);
            assertTrue(((PolygonMeshView) n).getMesh() instanceof PolygonMesh);
        }
```

AlmasB: ok, but second assert is redundant since `getMesh()` returns `PolygonMesh`

Issue #55, opened 2018-12-30
Publish to MavenCentral


jperedadnr:
So far artifacts are only published to Bintray

AlmasB:
Any ETA for this?

jperedadnr:
I've noticed I can publish to Maven Central from Bintray directly.

It just requires an extra step after `bintrayUpload`, publish to JCenter and sync to Maven Central, but it works. Artifacts are already available: https://search.maven.org/search?q=org.fxyz3d 
(importers will be ready tomorrow probably)

PR #75, opened 2019-02-09
#55 Don’t publish JavaFX classifier

jperedadnr:
(This is still work in progress to allow publishing to Maven Central)

This PR removes the classifier from the pom file of core/client/imports artifacts. This is required to avoid adding platform dependencies from where the library was published to other platforms that consume the library.

### DIFF
--- a/FXyz-Client/gradle/publishing.gradle
+++ b/FXyz-Client/gradle/publishing.gradle
@@ -100,11 +100,9 @@ publishing {
 
             pom.withXml {
                 def root = asNode()
-                root.dependencies.'*'.findAll() {
-                    it.groupId.text() == 'org.openjfx'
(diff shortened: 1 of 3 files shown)


AlmasB:
I can't comment on the content as I'm not familiar with gradle publishing. As for layout, for some reason I immediately read the following as `if ... else`:

```
                root.dependencies.'*'.findAll() {
                    it.groupId.text() == 'org.openjfx'
                }.each {
                    it.remove(it.classifier)
                }
```

If it's all the same, perhaps something like this is more similar to java streams layout?

```
root.dependencies.'*'.findAll() { it.groupId.text() == 'org.openjfx' }
                        .each { it.remove(it.classifier) }
```

PR #76, opened 2019-02-09
MayaImporter Test

AlmasB:
- correctly add maya view to Model3D (though distinct views can be separated still - future improvement)
- cleanup MayaImporter
- added RunWithFX extension for tests that require active JavaFX
- added MayaImporterTest
- added travis GUI emulation


FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
```diff
@@ -32,134 +32,103 @@
...
+
+        MayaGroup mayaRoot = new MayaGroup();
+        // Add top level nodes to the root
+        int nodeCount = 0;
+        for (Node n : loader.loaded.values()) {
+            if (n != null) {
+                // Only add a node if it has no parents, ie. top level node
+                if (n.getParent() == null) {
+                    if (Loader.DEBUG) {
+                        System.out.println("Adding top level node " + n.getId() + " to root!");
```
jperedadnr: Can we use a Logger instead? Or at least factorize it like in `ObjImporter`

AlmasB: I'm happy to change this to logger, all these debug statements seem to be all over the place

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
```diff
@@ -32,134 +32,103 @@
...
 
-            if (Loader.DEBUG) { System.out.println("Loaded " + count + " key frames."); }
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+        loader.keyFrameMap.forEach((seconds, keyValues) -> {
+            timeline.getKeyFrames().add(
+                    new KeyFrame(Duration.millis(seconds * 1000), keyValues.toArray(new KeyValue[0]))
+            );
+        });
+
```
nlisker: I would collect the `KeyFrame`s with:

    loader.keyFrameMap.entrySet().stream()
            .map(entry -> new KeyFrame(Duration.seconds(entry.getKey()), null, null, entry.getValue()))
            .forEach(timeline.getKeyFrames()::add);

* Stream allows us the future option of doing it in parallel, which, considering the amount of animation frames, seems relevant.
* `Duration.seconds` replaces `Duration.millis` and then multiplying by 1000.
* Using the `KeyFrame` constructor that takes a collection instead of an array saves the copying of `toArray`, which can save a lot of garbage. The 2nd and 3rd `null` arguments are the default. For more info see [JDK-8091817](https://bugs.openjdk.java.net/browse/JDK-8091817).

FXyz-Importers/src/test/java/org/fxyz3d/importers/maya/MayaImporterTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    void testLoadAsPoly() throws Exception {
+        Model3D model = importer.loadAsPoly(getClass().getResource("duke_king.ma"));
+
+        assertEquals(1, model.getRoot().getChildren().size());
+        assertEquals(1, model.getMeshNames().size());
+        assertTrue(model.getTimeline().isPresent());
+    }
```
jperedadnr: Adding a few more assertions wouldn't hurt:
```
assertEquals(1, model.getMeshViews().size());
        for (Node n : model.getMeshViews()) {
            assertTrue(n instanceof Group);
        }
```

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/MayaImporter.java
AlmasB: Great alternative.

FXyz-Importers/src/main/java/org/fxyz3d/importers/maya/values/impl/MArrayImpl.java
```diff
@@ -35,14 +35,12 @@
...
 import java.util.List;
-import org.fxyz3d.importers.maya.MayaImporter;
 import org.fxyz3d.importers.maya.types.MArrayType;
 import org.fxyz3d.importers.maya.values.MArray;
 import org.fxyz3d.importers.maya.values.MData;
 
 public class MArrayImpl extends MDataImpl implements MArray {
-    public static final boolean DEBUG = MayaImporter.DEBUG;
-    public static final boolean WARN = MayaImporter.WARN;
+    public static final boolean WARN = false;
```
jperedadnr: Should we remove this at all?

AlmasB: Sure, I'll see if there are any other places with similar patterns. I'll just replace with a Logger

AlmasB:
I've replaced sysout calls to Logger, but we do need to use a logging system, don't have a preference which one.

PR #78, opened 2019-02-10
Improve sampler controls for imported 3D models 

jperedadnr:


FXyz-Samples/src/main/java/org/fxyz3d/controls/FileLoadControl.java
```diff
@@ -29,26 +29,45 @@
...
+    private void chooseFile(ActionEvent event) {
+        FileChooser fileChooser = new FileChooser();
+        fileChooser.setTitle("Open Resource File");
+        File file = fileChooser.showOpenDialog(filePathTf.getScene().getWindow());
+        if (file != null) {
+            filePathTf.setText(file.getAbsolutePath());
+            try {
+                controlledProperty.setValue(file.toURI().toURL());
+            } catch (MalformedURLException e) {
+                e.printStackTrace();
```
AlmasB: Instead of `e.printStackTrace();` we really should have an app-wide exception handler. Something like this should suffice for now:

```
public class ExceptionHandler {
    public static void showError(Throwable e) {
        Alert alert = new Alert(Alert.AlertType.ERROR);
        alert.setTitle("Error occurred");
        alert.setContentText(e.toString());
        alert.show();
    }
}
```

FXyz-Samples/src/main/java/org/fxyz3d/controls/HierarchyControl.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public final void loadBindings() {
+        if (controlledProperty == null) {
+            return;
+        }
+        hierarchyTreeTable.rootProperty().bind(Bindings.createObjectBinding(() -> {
+                Node content3D = controlledProperty.getValue();
+                if (content3D != null) {
+                    return new TreeItemImpl(content3D);
```
AlmasB: Maybe `return (content3D != null) ? new TreeItemImpl(content3D) : null;`

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportSample.java
```diff
@@ -0,0 +1,173 @@
...
+                                model.sceneProperty().removeListener(this);
+                                timeline.set(null);
+                            }
+                        }
+                    });
+                    timeline.set(t);
+                }, () -> timeline.set(null));
+                initModel();
+            } catch (IOException e) {
+                e.printStackTrace();
```
AlmasB: Same here with exception handling.

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportOBJ.java
```diff
@@ -29,103 +29,25 @@
...
-        subdivision.addListener((obs, ov, nv) -> setSubdivisionLevel(model, nv.intValue()));
-    }
-    
-    private void setSubdivisionLevel(Node node, int subdivisionLevel) {
-        if (node instanceof PolygonMeshView) {
-            ((PolygonMeshView) node).setSubdivisionLevel(subdivisionLevel);
-        } else if (node instanceof Parent) {
-            ((Parent) node).getChildrenUnmodifiable().forEach(n -> setSubdivisionLevel(n, subdivisionLevel));
+        } catch (IOException e) {
+            e.printStackTrace();
```
AlmasB: Same here with exception handling

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportSample.java
```diff
@@ -0,0 +1,173 @@
...
+        });
+        culling.addListener((obs, b, b1) -> {
+            if (model != null) {
+                setCullFace(model, b1);
+            }
+        });
+        subdivision.addListener((obs, ov, nv) -> setSubdivisionLevel(model, nv.intValue()));
+        url.addListener((obs, ov, nv) -> {
+            try {
+                Model3D model3D;
```
AlmasB: `Model3D model3D = asPolygonMesh.get() ? Importer3D.loadAsPoly(url.get()) : Importer3D.load(url.get());`

FXyz-Samples/src/main/java/org/fxyz3d/controls/FileLoadControl.java
jperedadnr: Agree, but it is something for a new issue: logging and exception handler.

nlisker: Yes, we need proper levels logging for info/warning etc.

nlisker:
The hierarchy view on the right needs a lot more space both horizontally and vertically. Maybe make the right panel and the middle panel (model viewer) children of a split pane?

image

FXyz-Samples/src/main/java/org/fxyz3d/controls/HierarchyControl.java
```diff
@@ -0,0 +1,98 @@
...
+        if (controlledProperty == null) {
+            return;
+        }
+        hierarchyTreeTable.rootProperty().bind(Bindings.createObjectBinding(() -> {
+                Node content3D = controlledProperty.getValue();
+                return (content3D != null) ? new TreeItemImpl(content3D) : null;
+            }, controlledProperty));
+    }
+
+    private class TreeItemImpl extends TreeItem<Node> {
```
nlisker: Maybe `HierarchyTreeItem` or something like that is more descriptive than `TreeItemImpl`.

FXyz-Samples/src/main/java/org/fxyz3d/controls/HierarchyControl.java
```diff
@@ -0,0 +1,98 @@
...
+    @FXML private TreeTableColumn<Node, Boolean> visibilityColumn;
+
+
+    public HierarchyControl(Property<Node> prop) {
+        super("/org/fxyz3d/controls/HierarchyControl.fxml", prop);
+        nodeColumn.setCellValueFactory(p -> p.getValue().valueProperty().asString());
+        idColumn.setCellValueFactory(p -> p.getValue().getValue().idProperty());
+        visibilityColumn.setCellValueFactory(p -> p.getValue().getValue().visibleProperty());
+        visibilityColumn.setCellFactory(CheckBoxTreeTableCell.forTreeTableColumn(visibilityColumn));
+
```
nlisker: It would be nice to have a column for the material which is applied to the mesh/mesh group. Doesn't need to happen now. It could be a link to the material in a "Materials" titled pane or something like that, which we would want to allow the loaded materials to be inspected and edited, as we discussed in some comments in the past.

FXyz-Samples/src/main/java/org/fxyz3d/controls/HierarchyControl.java
```diff
@@ -0,0 +1,98 @@
...
+import javafx.scene.control.TreeTableView;
+import javafx.scene.control.cell.CheckBoxTreeTableCell;
+import org.fxyz3d.controls.ControlBase;
+
+/**
+ * FXML Controller class
+ *
+ * @author Jose Pereda
+ */
+public class HierarchyControl extends ControlBase<Property<Node>>{
```
nlisker: Space before `{`.

FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportSample.java
```diff
@@ -0,0 +1,171 @@
...
+                        }
+                    });
+                    timeline.set(t);
+                }, () -> timeline.set(null));
+                initModel();
+            } catch (IOException e) {
+                Logger.getLogger(ImportSample.class.getName()).log(Level.SEVERE,null, e);
+            }
+        });
+        yUp.addListener((obs, ov, nv) -> {
```
AlmasB: Can we bind the rotate angle value? (didn't check syntax)

```
Rotate rt = new Rotate(Rotate.X_AXIS);
rt.angleProperty().bind(Bindings.when(yUp).then(180).otherwise(0));
```

jperedadnr: Doing 
```  
model.setRotationAxis(Rotate.X_AXIS);
model.rotateProperty().bind(Bindings.when(yUp).then(180).otherwise(0));
````
kind of works, but it rotates the model 180º over its center and now, in case of the duke king, the model has its 0 in the top of the hat. Whereas the transform rotates over 0, and flips the model 180º, so the 0 still is at duke's feet.

Comment (AlmasB) [reply FXyz-Samples/src/main/java/org/fxyz3d/samples/importers/ImportSample.java]: I wasn't suggesting to bind the mode's angle, but to bind the `Rotate` transform's angle. Would that work? 

```
Rotate rt = new Rotate(Rotate.X_AXIS);
rt.angleProperty().bind(Bindings.when(yUp).then(180).otherwise(0));
model.getTransforms().setAll(rt);
```

jperedadnr: Yes, that works

nlisker:
Split pane looks good. A few more comments:

image

* The window is not resizable, it should be.
* The background looks really dark now compared to the previous image. It's hard to read the text.
* I don't understand what `@:` for the file dialog is.

PR #81, opened 2019-02-11
Initial MtlReader refactoring

nlisker:
Changes to `MtlReader`:

* Now reading with a stream, though it seems that mtl files are always short enough for sequential processing. Is this reasonable?
* Trying an `enum` approach as opposed to the `Map<Predicate, Consumer>`  approach in `ObjImporter` for the parser. Not sure if it's better or worse.
* Added missing properties, most of which are being ignored. Bump map and Specular map are treated now, but didn't test them.
* Now filtering empty lines and comments during reading. Do we really need to parse these like other lines? Same question goes for `ObjImporter`.
* What to do when reading a material which already exists? Completely replace it, ignore it, update by adding/overriding existing values, or throw an error?
* What to do when reading a property for a material which was already read? e.g.,
  ```
  newmtl mat
  Kd ...
  Ks ...
  Kd ... // again
  ```
  Replace, ignore, or throw an error?

### DIFF
--- a/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
+++ b/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
@@ -41,8 +41,10 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.stream.Stream;
 
@@ -71,6 +73,7 @@ public class MtlReader {
 
     private Map<String, Material> materials = new HashMap<>();
     private PhongMaterial currentMaterial;
+    private Set<String> readProperties = new HashSet<>(PARSERS.size() - 1);
 
     // mtl format spec: http://paulbourke.net/dataformats/mtl/
     private static final Map<String, BiConsumer<String, MtlReader>> PARSERS = Map.ofEntries(
@@ -98,14 +101,34 @@ public class MtlReader {
             entry("refl ",      (l, m) -> m.parseIgnore("Reflection map (refl)")),
             entry("map_aat ",   (l, m) -> m.parseIgnore("Anti-aliasing (map_aat)")));
 
+    private void parse2(String line) {
+        for (Entry<String, BiConsumer<String, MtlReader>> parser : PARSERS.entrySet()) {
+            String identifier = parser.getKey();
+            if (line.startsWith(identifier)) {
+                if (!"newmtl ".equals(identifier) && !readProperties.add(identifier)) {
+                    ObjImporter.log(identifier + "already read for current material. Ignoring.");
+                    return;
+                }
+                parser.getValue().accept(line.substring(identifier.length()), this);
+                return;
+            }
+        }
+        ObjImporter.log("No parser found for: " + line);
+    }
+
     private void parseIgnore(String nameAndKey) {
         ObjImporter.log(nameAndKey + " is not supported. Ignoring.");
     }
 
-    private void parseNewMaterial(String line) {
+    private void parseNewMaterial(String value) {
+        if (materials.containsKey(value)) {
+            ObjImporter.log(value + " material is already added. Ignoring.");
+            return;
+        }
         currentMaterial = new PhongMaterial();
-        materials.put(line, currentMaterial);
-//      TODO: ignore duplicates? ObjImporter.log("This material is already added. Ignoring " + line);
+        readProperties.clear();
+        materials.put(value, currentMaterial);
+        ObjImporter.log("Reading material " + value);
     }
     
     private void parseDiffuseReflectivity(String value) {
@@ -130,17 +153,6 @@ public class MtlReader {
 
     private void parseBumpMap(String value) {
         currentMaterial.setBumpMap(loadImage(value));
-    }
-
-    private void parse2(String line) {
-        for (Entry<String, BiConsumer<String, MtlReader>> parser : PARSERS.entrySet()) {
-            String identifier = parser.getKey();
-            if (line.startsWith(identifier)) {
-                parser.getValue().accept(line.substring(identifier.length()), this);
-                return;
-            }
-        }
-        ObjImporter.log("No parser found for: " + line);
     }
 
     private Color readColor(String line) {
--- a/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
+++ b/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
@@ -65,7 +65,7 @@ public class MtlReader {
         try (Stream<String> line = Files.lines(Paths.get(new URI(fileUrl)))) {
             line.map(String::trim)
                 .filter(l -> !l.isEmpty() && !l.startsWith("#"))
-                .forEach(this::parse2);
+                .forEach(this::parse);
         } catch (IOException | URISyntaxException e) {
             e.printStackTrace();
         }
@@ -101,7 +101,7 @@ public class MtlReader {
             entry("refl ",      (l, m) -> m.parseIgnore("Reflection map (refl)")),
             entry("map_aat ",   (l, m) -> m.parseIgnore("Anti-aliasing (map_aat)")));
 
-    private void parse2(String line) {
+    private void parse(String line) {
         for (Entry<String, BiConsumer<String, MtlReader>> parser : PARSERS.entrySet()) {
             String identifier = parser.getKey();
             if (line.startsWith(identifier)) {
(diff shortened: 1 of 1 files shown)


FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
```diff
@@ -54,87 +53,143 @@
     private String baseUrl;
 
     public MtlReader(String filename, String parentUrl) {
-        baseUrl = parentUrl.substring(0,parentUrl.lastIndexOf('/')+1);
+        baseUrl = parentUrl.substring(0, parentUrl.lastIndexOf('/') + 1);
         String fileUrl = baseUrl + filename;
-        try {
-            URL mtlUrl = new URL(fileUrl);
-            ObjImporter.log("Reading material from filename = " + mtlUrl);
```
jperedadnr: Can we keep the log?

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
```diff
@@ -54,87 +53,143 @@
...
-            read(mtlUrl.openStream());
-        } catch (FileNotFoundException ex) {
-            System.err.println("No material file found for obj. ["+fileUrl+"]");
-        } catch (IOException ex) {
-            ex.printStackTrace();
+        try (Stream<String> line = Files.lines(Paths.get(new URI(fileUrl)))) {
+            line.map(String::trim).filter(l -> !l.isEmpty() && !l.startsWith("#")).forEach(this::parse);
```
jperedadnr: One line is nice, of course, but I like each operation of the stream in one line

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
```diff
@@ -54,87 +53,143 @@
...
-                     */
-                } else {
-                    //log("material line ignored for " + name + ": " + line);
-                }
-            } catch (Exception ex) {
-                Logger.getLogger(MtlReader.class.getName()).log(Level.SEVERE, "Failed to parse line:" + line, ex);
+    private PhongMaterial currentMaterial;
+
+    // mtl format spec: http://paulbourke.net/dataformats/mtl/
+    private enum ParseKey {
```
jperedadnr: The enum looks ok, but why do we have two different approaches:  obj importer uses the Map with the predicates and biconsumers, mtl importer uses enums... shouldn't we use the same (I don't say which one of them)?

jperedadnr:
There is some issue when importing [this](http://hg.openjdk.java.net/duke/duke/file/e71b60779736/3D/Duke%20Waving) model: with the existing code, duke's body is white, with yours, it is black. 
If you add the missing texture file ([duke.png](http://hg.openjdk.java.net/duke/duke/file/e71b60779736/3D/Chess%20Duke/duke.png)), with the existing code it is added, but with yours, it is still black.

AlmasB:
> * Now reading with a stream, though it seems that mtl files are always short enough for sequential processing. Is this reasonable?

Given that parsing is stateful, I don't think we can parallelize this. This reminds me, I believe we can't parallelize ObjImporter either since parsing **has** to be done sequentially. So we can remove `isParallel`.

> * Added missing properties, most of which are being ignored. Bump map and Specular map are treated now, but didn't test them.

I wonder if we ever can make use of them. If not, perhaps we should just get rid of them for the sake of simplicity and print a warning message for each such property. 

> * Now filtering empty lines and comments during reading. Do we really need to parse these like other lines? Same question goes for `ObjImporter`.

As you suggested we should probably trim the lines first.

> * What to do when reading a material which already exists? Completely replace it, ignore it, update by adding/overriding existing values, or throw an error?

Imo we should ignore it and print a warning message to flag this up. After all, it must have been a flawed exporting since it doesn't make sense to have two materials with same name, right?

> * What to do when reading a property for a material which was already read? e.g.,

Same as above, read normally the first time round, if a property's been read, ignore and a warning message

nlisker:
> This reminds me, I believe we can't parallelize ObjImporter either since parsing **has** to be done sequentially. So we can remove `isParallel`.

Ah, I didn't know the order mattered for obj. Mtl could be parallelized by splitting into `newmtl` chunks and processing those in parallel, but it's useless to do so. Does obj support parallelizable chunks?

> I wonder if we ever can make use of them. If not, perhaps we should just get rid of them for the sake of simplicity and print a warning message for each such property.

Some of them we could make use of currently (for example, `d`/`Tr` could map to the alpha channel of the diffuse color maybe); some we could make use of in the future possibly (for example, adding ambient color to `PhongMaterial` or reflectivity to the graphics pipeline).

jperedadnr:
I don't think it's worth the effort: parallel won't work. 

Obj files, where parallel could make sense, are read sequentially, because there is no index assigned to each entity other than the one given by its current position in the file, relative to others elements of the same entity.

For Mtl, it doesn't make sense to do parallel over chunks of each material, as we are talking about a very limited number of properties.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
nlisker: We should. I wanted to consult to see if there are opinions towards one or the other. Honestly, even an `if-else` chain isn't *that* bad.
Part of the change is to remove the `line.substring(n)` job from the parsing function, where `n` is a magic number equal to the length of the key. I think I'm going back to the `Map` approach now that I thought about it a bit more and realized I can make it work with minor changes.

nlisker:
@jperedadnr I'm getting a 502 error on that page you linked. Can you make the files available somewhere?

> Second issue: the texture file is available, but when applied to the diffuse map, since you have set the diffuse color to black, you don't see the texture!

If the diffuse color is not set it is white. If it is set to black in the mtl file then that's not the importer's fault that the texture can't be seen. Why would setting a texture set the color too if the color was set to something else already?

jperedadnr:
Looks like https://hg.openjdk.java.net/ is down?
Alternative [link](https://www.dropbox.com/s/63ktqtq2xji1uva/duke_wave.zip?dl=0).

If the diffuse color is white, you are right, but there is a problem: The `material` instance is not reset, and inherits the color from the previous material. So setting it to white in this case fixes the issue.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
```diff
@@ -54,87 +59,100 @@
...
+            entry("map_Ks ",    (l, m) -> m.parseSpecularReflectivityMap(l)),
+            entry("map_Ns ",    (l, m) -> m.parseIgnore("Specular exponent map (map_Ns)")),
+            entry("map_d ",     (l, m) -> m.parseIgnore("Dissolve map (map_d)")),
+            entry("disp ",      (l, m) -> m.parseIgnore("Displacement map (disp)")),
+            entry("decal ",     (l, m) -> m.parseIgnore("Decal stencil map (decal)")),
+            entry("bump ",      (l, m) -> m.parseBumpMap(l)),
+            entry("refl ",      (l, m) -> m.parseIgnore("Reflection map (refl)")),
+            entry("map_aat ",   (l, m) -> m.parseIgnore("Anti-aliasing (map_aat)")));
+
+    private void parse2(String line) {
```
jperedadnr: the method can be just `parse`?

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/MtlReader.java
```diff
@@ -54,87 +59,100 @@
...
+    private void parseSpecularReflectivity(String value) {
+        currentMaterial.setSpecularColor(readColor(value));
+    }
+
+    private void parseSpecularExponent(String value) {
+        currentMaterial.setSpecularPower(Double.parseDouble(value));
+    }
+
+    private void parseDiffuseReflectivityMap(String value) {
+        currentMaterial.setDiffuseMap(loadImage(value));
```
jperedadnr: As discussed, we need to reset the material, or at least the diffuse color.

nlisker:
But the material is reset inside `parseNewMaterial` at `currentMaterial = new PhongMaterial();`. Looking at `Duke_wave.mtl`, this is the body material:

```
newmtl duke_w0286x_subdGeo_SG
illum 4
Kd 0.00 0.00 0.00
Ka 0.00 0.00 0.00
Tf 1.00 1.00 1.00
map_Kd duke.png
Ni 0.00
Ks 0.30 0.30 0.30
Ns 18.00
```

`Kd` is black, so `map_Kd` won't show. Set `Kd` to white and it shows.

jperedadnr:
Ok, I see. 
If you check the [specs](http://paulbourke.net/dataformats/mtl/):
> During rendering, the map_Kd value is multiplied by the Kd value.
if you set Kd = 0, it will always be black, no matter the texture. Probably an error in the file? I've imported it with Blender, and it is black too.

nlisker:
> During rendering, the map_Kd value is multiplied by the Kd value.

This is correct, and this is the behavior of the pixel/fragment shader in OpenJFX. In general, the ambient/ diffuse/specular color of a pixel is: (material color) * (texture color) * (light color), so if any is black the pixel is black.

I think the file is just wrong. `Kd` should be white.

PR #82, opened 2019-02-13
Refactor ObjImporter

nlisker:
Last refactor for now on `ObjImporter` so we can move to Maya.

* Removed stream parallel option and changed to multiline.
* Aligned the parsing mechanism with `MtlImporter` to remove the magic `substring` numbers when parsing.

### DIFF
--- a/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
+++ b/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
@@ -400,7 +400,7 @@ public class ObjImporter implements Importer {
         }
 
         private void parseSmoothGroup(String value) {
-            currentSmoothGroup = value.equals("off") ? 0 : Integer.parseInt(value.substring(2));
+            currentSmoothGroup = value.equals("off") ? 0 : Integer.parseInt(value);
         }
 
         private void parseMaterialLib(String value) {
--- a/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
+++ b/FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
@@ -149,7 +149,7 @@ public class ObjImporter implements Importer {
             for (Entry<String, BiConsumer<String, ObjModel>> parser : PARSERS.entrySet()) {
                 String identifier = parser.getKey();
                 if (line.startsWith(identifier)) {
-                    parser.getValue().accept(line.substring(identifier.length()), this);
+                    parser.getValue().accept(line.substring(identifier.length()).trim(), this);
                     return;
                 }
             }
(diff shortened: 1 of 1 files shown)


FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -134,22 +134,22 @@ private ObjModel read(URL url, boolean asPolygon) {
...
-                l -> l.startsWith("f "),                  (l, m) -> m.parseFace(l),
-                l -> l.startsWith("s "),                  (l, m) -> m.parseSmoothGroup(l),
-                l -> l.startsWith("mtllib "),             (l, m) -> m.parseMaterialLib(l),
-                l -> l.startsWith("usemtl "),             (l, m) -> m.parseUseMaterial(l),
-                l -> l.startsWith("vn "),                 (l, m) -> m.parseVertexNormal(l),
-                // comments and empty lines are ignored
-                l -> l.isEmpty() || l.startsWith("#"),    (l, m) -> {});
+        // obj format spec: http://paulbourke.net/dataformats/obj/
+        private static final Map<String, BiConsumer<String, ObjModel>> PARSERS = Map.of(
+                "g",       (l, m) -> m.parseGroupName(l),
```
jperedadnr: `l.startsWith("g ") || l.equals("g")` is now converted to `l.startsWith("g")`, which is not exactly the same. 
Is this refactoring really necessary? I know you don't like the use of `substring` ...

nlisker: The parser method was changed to take care of the group name. `g` is the identifier, and if there's something after it or not is not something the check condition should care about, but the job of the parser.

Refactoring is not necessary, but it helps. If the importer is improved with more parsing options it makes it a bit easier to do so.

jperedadnr: Ok, your approach works as the parser in this case is just taking everything that comes after `g`, hoping that there won't be other than these two cases:
`g`
`g <group name>`

In case some file contains something like:
`group name`
it will be parsed as "roup name", but either way that won't have any kind of repercusion.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
```diff
@@ -399,24 +399,24 @@ protected void parseFace(String line) {
             }
         }
 
-        private void parseSmoothGroup(String line) {
-            currentSmoothGroup = line.substring(2).equals("off") ? 0 : Integer.parseInt(line.substring(2));
+        private void parseSmoothGroup(String value) {
+            currentSmoothGroup = value.equals("off") ? 0 : Integer.parseInt(value.substring(2));
```
jperedadnr: `value.substring(2)` is wrong, should be just `value`.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
nlisker: `group` is not a valid start of line token. In fact, after `g` there can be more than 1 name separated by spaces, like
`g name1 name2`, which the parser doesn't treat.

I will also mention that the obj importer is incomplete in many other respects, and the mtl importer also doesn't account for all possibilities, from what I saw in the specs. The refactoring and cleanings of these will just make it easier to bring them up to specs in the future.

FXyz-Importers/src/main/java/org/fxyz3d/importers/obj/ObjImporter.java
nlisker: Good catch, I missed that one.

Issue #80, opened 2019-02-10
Option to visualize skeleton in skinningMesh


jperedadnr:
The skinningMesh has a tree of joints that defines all the transforms between the different elements.

 Adding a _bone_ node per joint with the proper transform allows showing the skeleton that was used to define the skinning mesh. 

<img width="1200" alt="skeleton" src="https://user-images.githubusercontent.com/2043230/52539081-55304e80-2d7a-11e9-9304-1634c6b8268e.png">

A boolean option can be added to visualize this skeleton when required.

Issue #52, opened 2018-11-05
Performance issues with animated Maya models


nlisker:
Taking the `King_WalkCycle.ma` as an example, loading and playing the animation of several models does not scale well. At 5 models the slowdown is noticeable and at 10 it's very jittery.

Garbage collection monitoring shows a lot of activity at 10 models with animations:

![image](https://user-images.githubusercontent.com/37422899/47976988-59cfd180-e0be-11e8-9a2d-e716eb38f8b4.png)

FalcoTheBold:
Because the entire armature is also loaded into the scene as a node + each joint being its own group-node.
Depending on the complexity of the armature this would rise the node-count dramatically of course.
Add to that the constant passing of event-objects when properties change, which may explain the high garbage collection.

Maybe a kind of arbitary armature-system, as in storing the raw data in a map accesible via joints as key, play animation as needed, etc., could be a better choice.
Currently the system is very clean, easily extendable and "working" though.

jperedadnr:
I've added 10 models of `King_WalkCycle.ma`, including the skeleton, and I see those peaks in the heap size:

<img width="553" alt="gc_kingwalk_x_10" src="https://user-images.githubusercontent.com/2043230/53241808-2338b980-36a3-11e9-8a68-8b920b951411.png">

What tool do you use to get the pauses?

nlisker:
Sorry, this issue fell off my radar unjustifiably.

I used JDK Mission Control. When the application is running, start a Flight Recorder recording, then switch to the Outline tab and under JVM Internals open Garbage Collection and you should see the same type of graph that I did.

Do you get a jittery animation? If not, try more models. The bottleneck should be in the same place in the code (not sure where yet), but maybe under a different load because of our different hardware.

PR #85, opened 2019-02-15
#80 Added JointChain that allows visualizing skeleton of a SkinningMesh

jperedadnr:
To test, run sampler, select Maya sample, go to 3D settings, and enable skeleton. Then play the animation.

### DIFF
--- a/FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/SkinningMesh.java
+++ b/FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/SkinningMesh.java
@@ -233,7 +233,7 @@ public class SkinningMesh extends PolygonMesh {
     }
 
     private void addJointChain(Joint joint) {
-        joint.getChildren().removeIf(JointChain.class::isInstance);
+        removeJointChain(joint);
         joint.getChildren().add(new JointChain(joint, 0.01));
     }
 
--- a/FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/SkinningMesh.java
+++ b/FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/SkinningMesh.java
@@ -86,12 +86,12 @@ public class SkinningMesh extends PolygonMesh {
         this.joints = joints;
 
         nJoints = joints.size();
-        nPoints = getPoints().size()/ getPointElementSize();
+        nPoints = getPoints().size() / getPointElementSize();
 
         // Create the jointIndexForest forest. Its structure is the same as
         // jointForest, except that this forest have indices information and
         // some branches are pruned if they don't contain joints.
-        jointIndexForest = new ArrayList<JointIndex>(jointForest.size());
+        jointIndexForest = new ArrayList<>(jointForest.size());
         for (Parent jointRoot : jointForest) {
             jointIndexForest.add(new JointIndex(jointRoot, joints.indexOf(jointRoot), joints));
         }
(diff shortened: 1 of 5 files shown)


FXyz-Core/src/main/java/org/fxyz3d/shapes/polygon/SkinningMesh.java
```diff
@@ -203,4 +207,38 @@ public void update() {
...
+    }
+    public final boolean isShowSkeleton() {
+       return showSkeleton.get();
+    }
+    public final void setShowSkeleton(boolean value) {
+        showSkeleton.set(value);
+    }
+
+    private void addJointChain(Joint joint) {
+        joint.getChildren().removeIf(JointChain.class::isInstance);
```
nlisker: Any reason not to use `removeJointChain` here to make it easy to see that adding a chain first removes the current one?

jperedadnr:
I can't reproduce the stuttering in my Mac. The line you mention updates the frustum mesh, so it probably is quite time and resource consuming. Not sure how we could improve this?

FXyz-Core/src/main/java/org/fxyz3d/utils/geom/JointChain.java
```diff
@@ -0,0 +1,141 @@
...
+    private void updateChain(Joint joint) {
+        Point3D scaled = getJointLocation(joint).multiply(1d / scale);
+        origin.setTranslateX(scaled.getX());
+        origin.setTranslateY(scaled.getY());
+        origin.setTranslateZ(scaled.getZ());
+        bone.setAxisOrigin(org.fxyz3d.geometry.Point3D.convertFromJavaFXPoint3D(scaled));
+        end.setTranslateX(bone.getAxisEnd().x);
+        end.setTranslateY(bone.getAxisEnd().y);
+        end.setTranslateZ(bone.getAxisEnd().z);
+
```
nlisker: Remove empty line.

nlisker:
I'm on Win10. I assume I will have to do performance analysis and see what the bottleneck is.

@AlmasB Do you get this stuttering too?

AlmasB:
Sorry I'm busy this week, will check at the weekend

nlisker:
Yes, looks like the problem is that updating the `FrustumMesh` calls the expensive `createFrustum`:

image

I'm not familiar with this class so I don't know why updating a property must create a new one. Also, why does the bone have to be a `FrustumMesh`, is a `Cylinder` not enough?

nlisker:
Seems like the issue persists (maybe not as badly though). I'll have to do profiling again, but only later today.

I wonder if this is somehow related to #52. Can you try to reproduce that one?

The hardware I'm on is i5-6500 and RX470 4GB if it helps with the comparison.

nlisker:
Looks like the issue is with `append` inside `getJointLocation`. If I do not append anything,

    Affine a = new Affine();
    for (int i = 0; i < 0; i++) {
        a.append(joint.getTransforms().get(i));
    }

there is no performance issue. If I change to `i < 1` it's stuttering. I'm looking into circumventing that chain, but the proper thing to do might be to create an MCVE and see if it's a bug in JavaFX.

nlisker:
By the way, I think you can merge the Cylinder revision for now if you want (I approved the PR since the performance issue is local). If you want to show the direction of the joint, maybe use a gradient texture?

Also, it might be a good idea to look into adding a cone, pyramid and maybe other basic shapes directly to JavaFX.

Did you have a chance to test #52? Could help with a direction.

jperedadnr:
I've merged it now as is. We shouldn't forget about the `append` issue though.

I've posted this [video](https://twitter.com/JPeredaDnr/status/1098912324889571328) btw, doing some tweaks to include a filled model and a wireframe model to view the skeleton...

Issue #84, opened 2019-02-14
Sampler window should be resizable


jperedadnr:
Also, there is an issue when maximizing it.

AlmasB:
@jperedadnr I've just skimmed through FXyzClient and it appears that the window is essentially `SimpleWindowFrame`. Is there a reason why we can't just use `Stage` itself?

jperedadnr:
Well, it is how it was designed initially (probably even from ControlsFX sampler), but there is also [commented out code](https://github.com/FXyz/FXyz/blob/master/FXyz-Client/src/main/java/org/fxyz3d/client/SimpleWindowFrame.java#L215) with a possible resizing implementation, that for some reason I can't recall was disabled.
Instead of changing the client, I'd rather reenable this code.

AlmasB:
Other than the amount of work needed to redesign the client to use the standard `Stage`, there is no strong reason to stay with current approach? In which case, I would be happy to redesign the client to use `Stage` directly. I think, ultimately, this redesign will lead to less code.

jperedadnr:
At the contrary, I don't see compelling reasons now to do a full redesign of the client. There are "historic" reasons to keep it as it, but of course, it is open to fixes and improvements. I'd rather spend time on the core, the importers, the samples...

PR #93, opened 2019-05-29
Sampler window is now resizable via south-east corner

AlmasB:
* Addresses the main issue in #84 
* Also added SE_RESIZE cursor when hovering over the resize area


FXyz-Client/src/main/java/org/fxyz3d/client/SimpleWindowFrame.java
```diff
@@ -213,25 +214,26 @@ private void initWindowControls() {
...
         });        
         southEastResize.setOnMouseDragged((e) -> {
-            double x = e.getScreenX() + dragOffsetX,
-                    y = e.getScreenY() + dragOffsetY;
+            double x = e.getScreenX() + dragOffsetX;
+            double y = e.getScreenY() + dragOffsetY;
             double w = x - getTranslateX();
             double h = y - getTranslateY();
 
             setPrefWidth(Math.max(stageMinWidth, w));
```
jperedadnr: The initial scene is set to 1200x768, but the actual min values are set to 1280x800, so resizing only works to larger values. 

It might be better to set min values like 800x600?

AlmasB: Good point. Will do that. Should I add resize areas at the bottom and the right sides?

jperedadnr: I guess SE is enough for now, but bottom and right will work too.

AlmasB: Ok might add S and E at a later time. Updated min values. Should be good to go.

Issue #98, opened 2019-07-17
CI: publish task fails because Sampler 


jperedadnr:
When releasing to Bintray, we just do `./gradlew bintrayUpload`, but that only can be applied to Core, Client and Imports, not to Sampler.

Fix should probably create a script to run three tasks:
```
./gradlew :FXyz-Core:bintrayUpload ...
./gradlew :FXyz-Client:bintrayUpload ...
./gradlew :FXyz-Importers:bintrayUpload ...
```

PR #100, opened 2019-08-07
Add jzy3d repository to pom

jperedadnr:


AlmasB:
Is travis error related to this PR?

jperedadnr:
Not directly to the pom changes, but to this change: https://github.com/FXyz/FXyz/pull/100/commits/53344324a1d18b77eff60faa225fc39bd30173dc
which is quite weird. I'll look into it

jperedadnr:
Ok, the build works now, but we have a different issue: `sh: 0: Can't open /etc/init.d/xvfb`
Based on this [question](https://stackoverflow.com/questions/55674746/travis-sh-0-cant-open-etc-init-d-xvfb), it looks like we have to modify travis.
I'll do it (there is also an issue with publishing all but samples, #98 )

jperedadnr:
Ok, we are green, we'll merge and do a minor release

PR #110, opened 2020-06-28
#109 Update Gradle, plugins and support JDK 14

### LINKED ISSUE #109: Upgrade to gradle 6.5 and support Java 14,15
This requires a new release of javafx-gradle-plugin: https://github.com/openjfx/javafx-gradle-plugin/issues/83

jperedadnr:
Fixes #109

### DIFF
--- a/FXyz-Client/build.gradle
+++ b/FXyz-Client/build.gradle
@@ -25,8 +25,6 @@ javadoc {
     failOnError = false
     
     options.windowTitle("FXSampler Project ${version}")
-    options.links("http://docs.oracle.com/javase/8/docs/api/");
-    options.links("http://docs.oracle.com/javase/8/javafx/api/");
     options.addBooleanOption("Xdoclint:none").setValue(true);
     options.addBooleanOption("javafx").setValue(true);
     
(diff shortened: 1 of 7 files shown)


nlisker:
I will review this sometime this week.

jperedadnr:
>All projects build (against JDK 12 and 14) with this change. 

Good

>However, the JUnit imports, such as org.junit.jupiter.api.extension.BeforeAllCallback, cannot be found. In Eclipse, I need to add a JUnit library to the module path and add requires entries in modules to get tests to compile.

This is happening now with this PR or from master too? Probably the issue was already there? `./gradlew test` works from command line, so maybe just an Eclipse issue?

> Also, Samples gives a deprecation warning about nashorn in module-info.java:44.

Yes this is a known issue. We have filed #59 from quite some time. We'll need to start investigating soon as this won't run with Java 15.

If there are no other issues, please approve this PR so we can integrate it.

nlisker:
> `./gradlew` test works from command line, so maybe just an Eclipse issue?

Probably. Gradle executes the tasks successfully. I will file an issue for Eclipse separately.

On another note, both before and after the patch I'm getting:
```
> Task :FXyz-Importers:test
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

PR #119, opened 2020-11-15
Organize build

### LINKED ISSUE #113: Organize dependencies
Importers/Client shouldn't depend on Core dependencies.

jperedadnr:
Fixes #113

Instead of having all dependencies in the parent build, move them to core. This allows, for instance, Importers to have only a clean dependency in the Core itself, and not in its transitive dependencies.

### DIFF
--- a/FXyz-Client/build.gradle
+++ b/FXyz-Client/build.gradle
@@ -16,10 +16,10 @@ if (!hasProperty('mainClass')) {
 }
 
 dependencies {    
-    implementation ('org.controlsfx:controlsfx:11.0.1') {
+    api ('org.controlsfx:controlsfx:11.0.2') {
         exclude group:"org.openjfx"
     }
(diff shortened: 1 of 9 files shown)


AlmasB (approved):

nlisker (approved):
I'm not too familiar with the build process, but I tested and it looks fine.

You can update gradle to 6.7 on the way.
Also, we might want to have version strings consolidated in some place instead of inline so it will be easier to update.

AlmasB:
Could we merge and release this soon? I wanted to use this as a dependency and need javafx upgraded to 15+

jperedadnr:
@nlisker I'd rather keep gradlew below 6.6 (see related [issue](https://github.com/openjfx/javafx-gradle-plugin/issues/89))

I'll file an issue about your suggestion.

PR #122, opened 2020-12-26
Refactor TorusMeshView

AlmasB:
- Added simple test
- Removed duplicated code
- ~~Renamed TorusMesh -> TorusMeshView~~ (since Mesh and MeshView are separate concepts, as defined by JavaFX itself). Given all `primitives` are currently named `Mesh`, I propose we rename all of them to correctly represent their nature.


jperedadnr:
Looks good. However, it would be better to have an (epic) issue first (i.e. rename all XXMesh primitives to XXMeshView), and discuss the scope of such change (i.e. how many primitives).

AlmasB:
Yeah, that makes sense, so I'll revert the `Mesh -> MeshView` rename for this one and we can deal with global renaming separately.

jperedadnr (approved):

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
```diff
@@ -1,353 +1,341 @@
...
+                        getTubeDivisions(),
+                        (float) getRadius(),
+                        (float) getTubeRadius(),
+                        (float) getTubeStartAngleOffset(),
+                        (float) getxOffset(),
+                        (float) getyOffset(),
+                        (float) getzOffset()
+                )
+        );
+    }
```
nlisker, line 89: Personally, I prefer the previous compact code style that avoids a lot of horizontal nesting, but I'm fine with this.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
```diff
@@ -1,353 +1,341 @@
...
+        int[] faces = new int[faceCount * 6];
+
+        int pointIndex = 0;
+        int texIndex = 0;
+        int faceIndex = 0;
+
+        float tubeFraction = 1.0f / tubeDivisions;
+        float radiusFraction = 1.0f / radiusDivisions;
+
+        float TWO_PI = 2.0f * 3.141592653589793f;
```
nlisker, line 115: It's easier to use `Math.PI` in the previous code to replace the explicit double value (`2 * Math.PI` instead of `2.0f * 3.141592653589793f`) than to create a special `TWO_PI` value, in my opinion.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
```diff
@@ -1,353 +1,341 @@
...
+                p1 = p1 % radiusDivisions != 0 ? p0 + 1 : p0 + 1 - radiusDivisions;
+
+                final int p0r = p0 + radiusDivisions;
+
+                final int p2 = p0r < numVerts ? p0r : p0r - numVerts;
+
+                int p3 = p2 < (numVerts - 1) ? p2 + 1 : p2 + 1 - numVerts;
+                p3 = p3 % radiusDivisions != 0 ? p2 + 1 : p2 + 1 - radiusDivisions;
+
+                try {
```
nlisker, line 157: What is this `try-catch` for?

nlisker (commented):
Additional comments:
1. Maybe the copyright years need to be updated to 2021 now.
2. I see that the faces creation algorithm changed and is much simpler now. I didn't review it. Did the previous code make superfluous calculations?
3. Do we put the properties code before or after the mesh code? In the `Spheroid` they come before, here they come after.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
AlmasB: Yes, but I didn't want to change the calculation itself, which used `float`.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
AlmasB: I don't know. I suspect index out of bounds. I didn't want to change it in this commit, but it will have to go.

AlmasB:
> 1. Maybe the copyright years need to be updated to 2021 now.
2. I see that the faces creation algorithm changed and is much simpler now. I didn't review it. Did the previous code make superfluous calculations?
3. Do we put the properties code before or after the mesh code? In the Spheroid they come before, here they come after.

1. Almost all changes, except for the revert in rename happened in 2020, so I guess it's fine
2. I just followed through with simple algebraic reduction and it appears yes previous code made superfluous calculations, or the formula being used just ended up giving same results in this specific context.
3. Good question, I'd like to delay this until all of the primitives have been refactored with 1-pass, I can then do another sweep with a more focused refactoring, to make code consistent. For this reason, I also didn't alter some of the code that can be cleaned up, e.g. the try-catch block.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
nlisker: Then maybe at least use `float TWO_PI = (float) (2 * Math.PI)`?

AlmasB: Sure.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
nlisker: Well, can the code reach a situation where an AIOOB can happen even? If so, it means that we need to do something about it in the `catch` block.

From what I see, the size of `faces` is `faceCount * 6 = (numVerts * 2) * 6 = (tubeDivisions * radiusDivisions)  * 12`. The loop runs over `tubeDivisions`  and `radiusDivisions`, and each iteration increments `faceIndex` by 12, so I don't think we can get out of bounds.

I would also move `int faceIndex = 0;` to just before the loops where it is used, but we can do it in another pass.

nlisker (approved):
Approving for a first-pass.

FXyz-Core/src/main/java/org/fxyz3d/shapes/primitives/TorusMesh.java
AlmasB: Yes, that was my plan for the future pass -- to see if AIOOB can happen.

PR #140, opened 2021-06-06
Remove Jzy3d dependency

### LINKED ISSUE #137: Remove jzy3d dependency
Up until 0.5.2 we had in the core pom:
```
<repositories>
   <repository>
    <id>jzy3d</id> 
    <url>http://maven.jzy3d.org/releases</url>
   </repository>
</repositories>
```
That is missing in 0.5.3, after the refactoring done in #130 

Either way, there is a major problem even if we add that back: gradle doesn't read the repository from the pom, and therefore it is required to include http://maven.jzy3d.org/releases as a repository.

To solve the need of such requirement, I propose to remove the Jzy3D dependency and include the original source code, which comes from https://github.com/yonatang/JDT/tree/master/core, within FXyz-Core.

jperedadnr:
Fixes #137

### DIFF
--- a/FXyz-Core/build.gradle
+++ b/FXyz-Core/build.gradle
@@ -2,15 +2,10 @@ ext.moduleName = 'org.fxyz3d.core'
 
 repositories {
     mavenCentral()
-    maven {
-        url = "http://maven.jzy3d.org/releases"
-    }
 }
 
 dependencies {
     api ('eu.mihosoft.vrl.jcsg:jcsg:0.5.7') { transitive = false }
     api 'eu.mihosoft.vvecmath:vvecmath:0.4.0'
     api 'org.orbisgis:poly2tri-core:0.1.2'
-    // A fork of JDT (Java Delaunay Triangulation) Core API from https://github.com/yonatang/JDT for Jzy3d
-    api 'org.jzy3d:jzy3d-jdt-core:1.0.2'
 }
--- a/FXyz-Core/src/main/java/module-info.java
+++ b/FXyz-Core/src/main/java/module-info.java
@@ -35,7 +35,6 @@ module org.fxyz3d.core {
     requires static jcsg;
     requires static vvecmath;
     requires static poly2tri.core;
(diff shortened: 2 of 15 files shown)


AlmasB (approved):
All looks good. Wonder if we need `IOParsers`.

nlisker:
I'm not sure why including the dependency just like we do in the other cases is a problem. Copying source code means it won't get updated. I'm not familiar with the code, so I can't comment on it.

jperedadnr:
@nlisker yes, I've been trying to keep the dependency to Jzy3d, but it is quite inconvenient, since it is not in Maven Central, it requires adding their repository http://maven.jzy3d.org/releases, as Gradle doesn't resolve it even if it is included in the the pom.
Anyway, Jzy3d is already using a copy of the original source code from https://github.com/yonatang/JDT/tree/master/core, which is 9 years old, and hasn't been updated since, and they published it, which was the reason we used their fork instead.
Does this make sense?

@AlmasB I'd rather keep the whole code from https://github.com/yonatang/JDT/tree/master/core, even if we don't use `IOParsers`.

nlisker:
If https://github.com/yonatang/JDT/tree/master/core is published can we not use it as a dependency?

jperedadnr:
If the author publishes it to Maven Central, yes, we definitely will use it as dependency. But I doubt that's going to happen, as the code is 9 years old, and it was never published before.

nlisker:
Oh, when you wrote "and they published it" I thought it was case.

If we can't solve it with dependencies then the patch if fine.

jperedadnr:
Sorry, I was not clear. 

- Original code: https://github.com/yonatang/JDT/tree/master/core (never published)
- Jzy3d forked it into https://github.com/jzy3d/jzy3d-api/tree/master/jzy3d-jdt-core, and published it to http://maven.jzy3d.org/releases, which is what we have been using as dependency.

This PR removes the dependency from the Jzy3d fork, and adds the source code, so no external dependencies are required.

nlisker (approved):

Issue #136, opened 2021-04-20
OBJ importer does not recognize object identifiers


armin-reichert:
For my (Ms.) Pac-Man game implementation, I use a 3D model created using Blender 2.92 (by some nice guy, not by myself) and exported into OBJ format. See here:

https://github.com/armin-reichert/pacman-javafx/tree/main/pacman-ui-fx/src/main/resources/common/gianmarco

Currently I am importing this OBJ file using the importer by InteractiveMesh.org:

http://www.interactivemesh.org/models/jfx3dimporter.html

As there is no Maven artifact for that importer, I tried to use the FXyz importer but after importing the model, all object identifiers from the OBJ file are missing. Instead there appear generic identifiers like "default (3)" all over the place. 

A quick look into the parser revealed that lines in the OBJ file starting with "o " are not parsed as identifiers. See class ObjImporter.ObjModel.

I assume it will be sufficient to add a key-value pair for the lines starting with "o " into the map PARSERS in class ObjImporter.ObjModel and provide the code to read the rest of the line into the identifier key.

I tried to fix that issue locally, cloned the FXyz repository and imported the Gradle project into my Eclipse IDE. Here I get errors concerning the JUnit tests because the Gradle dependencies to JUnit5 seems to be incomplete. I have no experience at all with Gradle and cannot fix that quickly. 

So these are infact two issues of which only the importer issue is important (sic) to me.  

Thanks!

armin-reichert:
I managed to get this running in my Eclipse IDE and fixed the bug in the parser.

jperedadnr:
Do you want to contribute the fix?

armin-reichert:
Hi Jose,

my original intention was to use the FXyz importer instead of the one I currently use (interactivemesh.org) because of the missing Maven support of the latter. When I noticed that the FXyz importer ignores "o" tags, I "fixed" that in a local copy. But even after this fix the imported mesh and material maps still differ between the two importers. For this reason I did not send a pull request but stayed with my other importer. 

So I would rather see a fix made by someone who is more familiar with the FXyz importer than I am.

Regards
Armin

jperedadnr:
You can choose the importer that best fits of course, but would be good to see why they are different, as that means that one of them is doing it wrong.

In the pacman case, I see that `o` could be replaced with `g`, as there are no groups part of objects. What other differences do you see?

armin-reichert:
[...]

PR #144, opened 2021-06-07
Correctly store (and reuse) material data for Model3D

AlmasB:
This is the first step towards addressing #136.

Ideally we should have a concept of an `object - o` or a `group - g`, as defined by the obj file format, so we can address #136. Then we won't need to map mesh names to material names. After the first pass, we would produce one or more `objects` (or `groups`), which can then be used for generating mesh views. However, at this time, the introduced mapping simplifies the PR by minimising the number of changes.


nlisker:
I think we can just implement the concepts of groups (which we more or less have) and object names to begin with. I'm looking at http://paulbourke.net/dataformats/obj/ and it doesn't give too much information on what to do with them, only on how to read them.

We can add a field `String objectName` and when we encounter an `o` statement we assign its value to the field. Then, until we encounter a new definition, we apply the name to the next values. We can keep a map from the `String`s to the objects for example, but that depends on what we want to do with it.

Birdasaur:
@jperedadnr @nlisker I see review is passed. Is there any reason to wait on merging? Sorry this has been sitting here so long.

Issue #133, opened 2021-02-27
[Solved] Skybox memory leak


ivanstepanovftw:
Solved by commenting `startTimer();` in Skybox.java

jperedadnr:
Can you clarify the issue and how to reproduce it?

ivanstepanovftw:
> PhD Structural Engineer, Java Champion, Oracle Dev Champion. Love coding Java & JavaFX, mobile apps, IoT. Software Engineer at http://gluonhq.com/

ivanstepanovftw:
```
for (i in 1..10000) {
node.getChildren().add(new Skybox());
}
```

ivanstepanovftw:
Not solved, why are you closing https://github.com/FXyz/FXyz/blob/master/FXyz-Core/src/main/java/org/fxyz3d/scene/Skybox.java

ivanstepanovftw:
Add method to stop the timer or remove it completely

Birdasaur:
Or you submit a PR to add the additional functionality you desire.
Last I checked we don't work for you.

Issue #104, opened 2020-03-01
Bug in Skybox.java


wholder:
These two lines in `Skybox.java` have the back and front images reversed:

```
        back.setImage(frontImg);
        front.setImage(backImg);

```
It should be:

```
        back.setImage(backImg);
        front.setImage(frontImg);

```
Wayne

jdub1581:
It has been quite a while since I have been here.. Appologies to the Community for my abscence and any possible transgressions. I'll take a look, I believe it was the orientation of this particular image. I probably did not refactor it for that reason.

PR #167, opened 2023-01-13
Fix two small skybox issues

Birdasaur:
Addresses issues https://github.com/FXyz/FXyz/issues/104 and https://github.com/FXyz/FXyz/issues/133

Fixed Double boxing deprecation warning.

### DIFF
--- a/FXyz-Core/src/main/java/org/fxyz3d/scene/Skybox.java
+++ b/FXyz-Core/src/main/java/org/fxyz3d/scene/Skybox.java
@@ -1,7 +1,7 @@
 /**
  * Skybox.java
  *
- * Copyright (c) 2013-2016, F(X)yz
+ * Copyright (c) 2013-2023, F(X)yz
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


FXyz-Core/src/main/java/org/fxyz3d/scene/Skybox.java
```diff
@@ -121,7 +121,13 @@ public Skybox(Image topImg, Image bottomImg, Image leftImg, Image rightImg, Imag
         startTimer();
     }
 
-    
+    public void setEnableTimer(boolean enable) {
```
jperedadnr, line 124: I take this is just new API to allow developers access the timer? Developers can use it to start/stop the timer, but there is no internal handling of this?

Do you think the samples that use the skyBox could benefit from this API?

For instance, would make sense do have a listener to the scene property of the skyBox, so whenever it is part of the scene, timer starts, and when removed it stops?

We don't have javadoc (yet), but maybe we can add here:
```
/**
 * @since 0.5.5
 */
```?

Birdasaur: Yes only intent is to give Developers full control. Crybaby in #133 seemed to have a use case for rapidly adding and removing Skybox's from the scene.

I'm on the fence for adding a listener since it adds behavior that is hidden by default. I suppose a compromise would be adding a bindable property which gives the developer an option via binding to automatically start/stop? 

Ahh javadoc... you're going to make me do this the right way aren't you? :-)

Birdasaur: Ok so for updating the samples... you're suggesting updating the SkyBox sample with some control that uses the new public API call right?

Birdasaur: I think we need a follow up on any potential sample changes simply because I dont see what controls we could add currently.

jperedadnr (approved):
Looks good!

PR #171, opened 2023-08-18
Remove thirdPartyCompatibility from gradle module metadata

### LINKED ISSUE #174: Gradle Module Metadata locked in to Linux Classifier
The gradle module metadata (gmm) upstream in central seems to reference `linux` classifier on the JFX dependencies. [See example from central](https://repo1.maven.org/maven2/org/fxyz3d/fxyz3d/0.5.4/fxyz3d-0.5.4.module). This causes issues, such as when attemting to jlink (as a consumer of the library) on gradle which will make the `-linux` classifier version to be pulled in and result in errors like:
 `Error: Two versions of module javafx.base found in /path/to/build/jlinkbase/jlinkjars (javafx-base-17.0.2-mac-aarch64.jar and javafx-base-17.0.2-linux.jar)`

samypr100:
Closes #174

The gradle module metadata (gmm) upstream in central seems to reference `linux` classifier on the JFX dependencies. [See example from central](https://repo1.maven.org/maven2/org/fxyz3d/fxyz3d/0.5.4/fxyz3d-0.5.4.module). This causes issues, such as when attemting to jlink (as a consumer of the library) on gradle which will make the `-linux` classifier version to be pulled in and result in errors like:
 `Error: Two versions of module javafx.base found in /path/to/build/jlinkbase/jlinkjars (javafx-base-17.0.2-mac-aarch64.jar and javafx-base-17.0.2-linux.jar)`

The added code snipped does the equivalent to `module.json` that this does on the `pom.xml` in `publishing.gradle`

```groovy
pom.withXml {
    def root = asNode()
    root.dependencies.'*'
            .findAll() { it.groupId.text() == 'org.openjfx' }
            .each { it.remove(it.classifier) }
    root.appendNode 'description', publish.desc
    root.children().last() + pomConfig
}
```

This will help with [Trinity](https://github.com/Birdasaur/Trinity) updates @Birdasaur

Edit: This PR now targets the new JavaFX Plugin Updates which helps resolves the `thirdPartyCompatibility ` issue. The code snippet has been removed, and the pom changes have been updated since we don't need to remove the classifier any longer.


jperedadnr:
I wasn't aware of this Gradle module metadata, that creates a parallel pom of this dependency just for Gradle projects... 

More can be found [here](https://github.com/gradle/gradle/blob/master/subprojects/docs/src/docs/design/gradle-module-metadata-latest-specification.md).

I see that the same we did for the pom (removing the classifier) has to be done for this .module file. 

From what I get from the above doc:

> `thirdPartyCompatibility`: optional. Includes additional information to be used if the dependency points at a module that did not publish Gradle module metadata.
> These are typically artifacts published with Maven that contain a _classifier_.

If I take it right, it is an optional field that can be excluded without any risk, since the traversal dependency from the base jar to the classifier jar should be resolved elsewhere (i.e. javafx plugin, or jlink task)?

jperedadnr:
@abhinayagarwal added you to review this PR, just in case we need to apply the same to other JavaFX/gradle repositories.

jperedadnr:
Is this a more simple option? Do we really need GMM?

```
// Do not publish Gradle modules
    tasks.withType(GenerateModuleMetadata) {
        enabled = false
    }
```

samypr100:
Good point, disabling it all together is another valid approach. Since you have packages published with it already I opted to keep it to avoid affecting downstream consumers. I have a gradle feature request open so they add a `thirdPartyCompatibility` flag that allows to remove that section all together from the generated module.json since it's optional to begin with per the spec.

samypr100:
> If I take it right, it is an optional field that can be excluded without any risk, since the traversal dependency from the base jar to the classifier jar should be resolved elsewhere (i.e. javafx plugin, or jlink task)?

Correct, although upstream JFX [doesn't publish](https://repo1.maven.org/maven2/org/openjfx/javafx-base/20.0.2/) a GMM yet. The way it stands gradle will use [artifact()](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/repositories/MavenArtifactRepository.MetadataSources.html) as a fallback to get the classifier jars assuming there's no GMM to indicate otherwise (like the case of FXyz). Long term ideal solution might be for upstream jfx to publish a GMM that contains all the classifiers.

jperedadnr (approved):
I'm fine with this for the time being. When there is a better fix, we can revisit this.

samypr100:
@jperedadnr this PR might be able to be closed if this gets merged https://github.com/openjfx/javafx-gradle-plugin/pull/151

samypr100:
@jperedadnr This PR is now moved to use the `0.0.15-SNAPSHOT` update

build.gradle
```diff
@@ -5,7 +5,8 @@ plugins {
     id 'net.nemerosa.versioning' version '3.0.0'
     id 'com.github.ben-manes.versions' version '0.44.0'
     id 'io.codearte.nexus-staging' version '0.30.0'
-    id 'org.openjfx.javafxplugin' version '0.0.13'
+    id 'org.openjfx.javafxplugin' version '0.1.0' apply false
+    id 'org.javamodularity.moduleplugin' version '1.8.12' apply false
```
jperedadnr, line 9: Why do we still need `org.javamodularity.moduleplugin`?

jperedadnr (commented):
Can you also update the title of the PR?

build.gradle
jperedadnr: If it is because automatic modules are not recognised with gradle 8.3, we might consider `org.gradlex:extra-java-module-info:1.4.1` instead?

samypr100: It was transitively depending on it via the javafx plugin previously

samypr100:
> Can you also update the title of the PR?

Thought on a name here? The PR is still about getting rid of thirdPartyCompatibility, which now updating the plugin accomplishes that 😅

build.gradle
samypr100: We can switch to using `org.gradlex:extra-java-module-info:1.4.2` if desired, although I haven't used that plugin before.

I believe these would be the minimal amount of changes (explicitly declaring all legacy jars as automatic modules)

```gradle
// build.gradle
// We can remove modularity plugin
java {
    modularity.inferModulePath.set(true)
}
...
// FXyz-Samples/build.gradle
extraJavaModuleInfo {
    automaticModule("eu.mihosoft.vrl.jcsg:jcsg", "jcsg")
    automaticModule("eu.mihosoft.vvecmath:vvecmath", "vvecmath")
    automaticModule("org.orbisgis:poly2tri-core", "poly2tri.core")
    automaticModule("org.slf4j:slf4j-api", "slf4j.api")
    automaticModule("org.reactfx:reactfx", "reactfx")
    automaticModule("org.fxmisc.easybind:easybind", "easybind")
}
...
// FXyz-Client/build.gradle
extraJavaModuleInfo {
    automaticModule("org.reactfx:reactfx", "reactfx")
}
...
// FXyz-Core/build.gradle
extraJavaModuleInfo {
    automaticModule("eu.mihosoft.vrl.jcsg:jcsg", "jcsg")
    automaticModule("eu.mihosoft.vvecmath:vvecmath", "vvecmath")
    automaticModule("org.orbisgis:poly2tri-core", "poly2tri.core")
    automaticModule("org.slf4j:slf4j-api", "slf4j.api")
}
```

jperedadnr: The gradlex plugin is from @jjohannes, so I guess this is the way to go, in line with the changes in the JavaFX gradle plugin.

jjohannes: Some notes:
- Automatic-Modules are recognized by Gradle, if there is an "Automatic-Module-Name" entry in the Manifest. So I am not sure if you would need all the things listed above. E.g. slf4j is a "real" module in the latest version. If you run the build with the plugin without defining any patching, it will tell you which Jars have to be patched.
- If you use the plugin and do patching in a library (not an application) build like this one, you should be carful which Module Names you choose if the names also turn up int your `module-info.java`. Because if you choose a different one as the one that Java would give based on the Jar name, users of the library would have to do the same patching. For example, in the list about "slf4j.api" could be problematic: It's true if the Jar has no automatic module name in the version you are using, but the official name in the current version is "org.slf4j" (without api). 

It is still a very messy situation right now, especially for libraries that have transitive dependencies which are not yet completely modularized. :( I think I would do this in an separate PR and continue to use the other plugin for now if that is what was done implicitly before.
Happy to give more pointers or do a review if you do the change in another PR.

samypr100: Thanks great point on slf4j naming, I didn't even think about it. I'm hopeful we can remove the slf4j entries when we merge https://github.com/FXyz/FXyz/pull/173 which upgrades it to 2.x.
All the ones added above were ones that the plugin errored on via `Not a module and no mapping defined: xyz.jar`

I also just opened https://github.com/openjfx/javafx-gradle-plugin/pull/154 to update docs on the removal of `org.javamodularity.moduleplugin` for others that might encounter this situation.

jperedadnr: #173 merged.
@samypr100 Do you want to update this PR? Do we merge this one before #172?

samypr100: I think we'll want to merge this one last of the 3

samypr100: We can switch to use the org.gradlex:extra-java-module-info plugin if you're alright with that or I'd be okay doing that on a different PR. Whichever you prefer.

jperedadnr: Ok, let's go for now with your current approach, and we'll switch plugins if needed.

PR #172, opened 2023-08-18
Upgrade Gradle to 8.3

samypr100:
Closes https://github.com/FXyz/FXyz/issues/175

Used `./gradlew wrapper --gradle-version 8.2.1` to update Gradle to 8.2.1.

Updated some plugins and resolved some deprecations that were showing up due via  `--warning-mode all`

Last set of deprecations would involve using https://github.com/gradle-nexus/publish-plugin and removing `de.marcphilipp.nexus-publish` and `io.codearte.nexus-staging`, but that can be a separate PR

Edit (Sept 2nd 2023): Updated to Gradle 8.3 in 819280799afc1a2d69de03a6988d587109153d41


jperedadnr:
Bumping the gradle version is always scary... This needs some testing, so I'm adding extra reviewers in case any of them has the chance to test this PR, though just two approvals will be needed.

jperedadnr (approved):
Looks good to me. Tested locally, no major issues.

samypr100:
> Looks good to me. Tested locally, no major issues.

@jperedadnr I've bumped the PR to target 8.3

jperedadnr:
Okay, that’s good,.
As you might be aware, there will be a new JavaFX plugin release with some changes, based on this PR:  https://github.com/openjfx/javafx-gradle-plugin/pull/151 that was just integrated. 
Is it possible that you can test the snapshot?

samypr100:
Agreed, I'll push those to the other PR and mark it as draft to keep this one clean to just the gradle upgrade

nlisker:
I think we should add to the build file:
```
wrapper {
    gradleVersion = gradleWrapperVersion
}
```
where `gradleWrapperVersion` is the version that we update. Running this will produce the wrapper files so that we won't have to update those instead. It can be defined as
```
ext.gradleWrapperVersion = "8.3";
```

samypr100:
> I think we should add to the build file:
> 
> ```
> wrapper {
>     gradleVersion = gradleWrapperVersion
> }
> ```
> 
> where `gradleWrapperVersion` is the version that we update. Running this will produce the wrapper files so that we won't have to update those instead. It can be defined as
> 
> ```
> ext.gradleWrapperVersion = "8.3";
> ```

@nlisker Added in 3af4d1a41e454217afa71ee6cf750d341d2f689a

build.gradle
```diff
@@ -21,9 +22,11 @@ subprojects {
     apply plugin: 'com.github.hierynomus.license'
 
     java {
+        sourceCompatibility = JavaVersion.VERSION_11
+        targetCompatibility = JavaVersion.VERSION_11
```
nlisker, line 26: I think it would be better to use the [`--release`](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.compile.CompileOptions.html#org.gradle.api.tasks.compile.CompileOptions:release) option.

samypr100: I kept it the same as before to avoid out of scope changes and keep it minimal (e.g. switching to release is not needed for the gradle 8.3 update).
On that topic I'd recommend moving to [toolchains](https://blog.gradle.org/java-toolchains) instead when pivoting away from these options.

nlisker: That would be my preference as well in general. We'll do it in a followup issue.

jperedadnr: @nlisker Are you okay with this PR as is then?

nlisker: Yes

Issue #87, opened 2019-03-06
Hotfix to exclude slf4j-simple supporting Java 8


calvertdw:
We use Log4j2 and get the following annoyance when running with fxyz3d 0.1.1 on our classpath.

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/C:/Users/Duncan/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-slf4j-impl/2.11.1/4b41b53a3a2d299ce381a69d165381ca19f62912/log4j-slf4j-impl-2.11.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/C:/Users/Duncan/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-simple/1.6.1/58e59bfb3e247097b8122243b3bfe0049c8cfae8/slf4j-simple-1.6.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory]
```
The transitive comes from `jcsg`.
![image](https://user-images.githubusercontent.com/1161586/53846790-e9ee2c80-3f73-11e9-94f1-b21eac55ce4b.png)

We currently still use Java 8, so it appears a hotfix to 0.3.0 would be appropriate. Would this be something you guys could do? I'm also open to forking and releasing the hotfix myself.

Tagging @SylvainBertrand

jperedadnr:
Actually FXyz has two dependencies in `eu.mihosoft.vrl.jcsg:jcsg` and `org.orbisgis:poly2tri-core`, and both use `org.slf4j`.

I see that you [use](https://github.com/ihmcrobotics/ihmc-javafx-toolkit/blob/758360ad5a1cc5248e9425c502cc0b11ffd84b2d/src/main/java/us/ihmc/javaFXToolkit/text/Text3D.java) the `Text3D` shape from FXyz, which relies on Poly2tri, which at the end has the dependencies on slf4j (see for instance [Polygon](https://github.com/orbisgis/poly2tri.java/blob/master/poly2tri-core/src/main/java/org/poly2tri/geometry/polygon/Polygon.java#L10)).

So this is something we can't change.

However if you do something like:
```
[...]

PR #173, opened 2023-08-18
Update to slf4j 2.x for jpms compatibility

samypr100:
Closes https://github.com/FXyz/FXyz/issues/176

Somewhat related to https://github.com/FXyz/FXyz/issues/87

This upgrades slf4j to mark it as a runtime dependency, and also to 2.x which is jpms compatible and allows it to work smoothly with pure JLink downstream. Used `requires` instead of `requires transitive` on `org.slf4j` in `module-info.java` per [gradle docs conventions](https://docs.gradle.org/current/userguide/java_library_plugin.html#declaring_module_dependencies).

This will help with [Trinity](https://github.com/Birdasaur/Trinity) updates @Birdasaur

### DIFF
--- a/FXyz-Core/build.gradle
+++ b/FXyz-Core/build.gradle
@@ -7,5 +7,11 @@ repositories {
 dependencies {
     api ('eu.mihosoft.vrl.jcsg:jcsg:0.5.7') { transitive = false }
     api 'eu.mihosoft.vvecmath:vvecmath:0.4.0'
-    api 'org.orbisgis:poly2tri-core:0.1.2'
+    api ('org.orbisgis:poly2tri-core:0.1.2') {
+        exclude group: 'org.slf4j', module: 'slf4j-api'
+    }
+    // Both eu.mihosoft.vrl.jcsg and org.orbisgis:poly2tri-core depend on this.
+    // Using 2.x instead to have a JPMS compatible version and "implementation"
+    // configuration to make it a runtimeElement when publishing.
+    implementation 'org.slf4j:slf4j-api:2.0.7'
 }
--- a/FXyz-Core/build.gradle
+++ b/FXyz-Core/build.gradle
@@ -13,5 +13,5 @@ dependencies {
     // Both eu.mihosoft.vrl.jcsg and org.orbisgis:poly2tri-core depend on this.
     // Using 2.x instead to have a JPMS compatible version and "implementation"
     // configuration to make it a runtimeElement when publishing.
-    implementation 'org.slf4j:slf4j-api:2.0.7'
+    implementation "org.slf4j:slf4j-api:$slf4j_version"
 }
--- a/FXyz-Core/src/main/java/module-info.java
+++ b/FXyz-Core/src/main/java/module-info.java
@@ -25,7 +25,7 @@
  * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */ 
+ */
 
 module org.fxyz3d.core {
     requires transitive javafx.controls;
@@ -35,6 +35,7 @@ module org.fxyz3d.core {
     requires static jcsg;
     requires static vvecmath;
     requires static poly2tri.core;
+    requires org.slf4j;
 
     exports org.fxyz3d.geometry;
     exports org.fxyz3d.io;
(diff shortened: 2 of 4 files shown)


jperedadnr:
@samypr100 checks are failing. Can you see what is the problem about?

FXyz-Core/build.gradle
```diff
@@ -7,5 +7,11 @@ repositories {
 dependencies {
     api ('eu.mihosoft.vrl.jcsg:jcsg:0.5.7') { transitive = false }
     api 'eu.mihosoft.vvecmath:vvecmath:0.4.0'
-    api 'org.orbisgis:poly2tri-core:0.1.2'
-}
\ No newline at end of file
+    api ('org.orbisgis:poly2tri-core:0.1.2') {
+        exclude group: 'org.slf4j', module: 'slf4j-api'
```
jperedadnr, line 11: Since you are removing this api dependency from FXyz-Core, FXyz-Imports, that depends on FXyz-Core fails to find it.
So it can be fixed adding the same dependency as implementation for FXyz-Imports. Does that make sense to you?

FXyz-Core/src/main/java/module-info.java
```diff
@@ -25,7 +25,7 @@
  * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */ 
+ */
```
jperedadnr, line 28: Update license header:
```
Copyright (c) 2013-2023, F(X)yz
```

FXyz-Core/build.gradle
samypr100: Sorry about that, I was testing samples jlink and forgot to run tests on this one haha.

I just realized importers had `api (project(':FXyz-Core')) { transitive = false }` transitives set to false.

I updated it so that the runtimeElements have the dependencies, hence the second `implementation (project(':FXyz-Core'))`

This keeps apiElements metadata correct. So there's no changes downstream.

samypr100: > So it can be fixed adding the same dependency as implementation for FXyz-Imports. Does that make sense to you?

Sadly it would've failed since transitives is set to false. I would've needed to change module-info.java to use `requires static org.slf4j` to make that work but then jlink would not be happy since slf4j would not an automatic module anymore.

FXyz-Core/src/main/java/module-info.java
samypr100: Updated, thx

FXyz-Importers/build.gradle
```diff
@@ -2,4 +2,5 @@ ext.moduleName = 'org.fxyz3d.importers'
 
 dependencies {
     api (project(':FXyz-Core')) { transitive = false }
-}
\ No newline at end of file
+    implementation (project(':FXyz-Core'))
```
jperedadnr, line 5: Having both api and implementation for the core project doesn't make sense, imo.

Would this work:
```
dependencies {
    api (project(':FXyz-Core')) { transitive = false }
    runtimeOnly 'org.slf4j:slf4j-api:2.0.7'
}
```
or:
```
dependencies {
    api (project(':FXyz-Core')) { transitive = false }
    implementation 'org.slf4j:slf4j-api:2.0.7'
}
```
?

samypr100: Yes, I was trying to re-use the `project` to avoid the overhead to keep the versions in sync in two places.

You can also use this to keep the the GMM exactly the same for runtimeElements as well.

```
dependencies {
    api (project(':FXyz-Core')) { transitive = false }
    runtimeOnly (project(':FXyz-Core'))
}
```

samypr100: This is also a possibility since slf4j is only needed in the importers tests

````grooby
dependencies {
    api (project(':FXyz-Core')) { transitive = false }
    testRuntimeOnly (project(':FXyz-Core'))
}

jperedadnr: Yes, that's better.

In any case, just for using one dependency of the whole core project, I'm still in favour of including only such dependency, and look for a better way to find its version (for instance with a variable in the parent gradle file that can be reused by core and imports).

samypr100: Ah I see. If you support using properties here, I'll do so :)

samypr100: See 910e8807a44afdcf23eda154108b9162ecb2e9e4

jperedadnr (approved):

jperedadnr (approved):
Looks good

PR #179, opened 2023-09-06
Obj: removed silent ignore of errors

ennerf:
`ObjImporter::load` is marked to throw an `IOException`, but the internal `ObjImporter::read` silently ignores any errors and never throws one. Invalid urls and missing files are happily parsed with no way to verify the result.

This PR passes the Exception up so users can be notified if e.g. a URL is not available.


jperedadnr (commented):
Looks good

Issue #180, opened 2023-09-07
Move to extra-java-module-info plugin


samypr100:
Per discussion in https://github.com/FXyz/FXyz/pull/171#discussion_r1315297483 this issue aims to capture
* Moving to use the `extra-java-module-info` and define missing module information for "legacy" jars.
* Removing the `org.javamodularity.moduleplugin` plugin.
* Setting `java.modularity.inferModulePath` back to `true`.

nlisker:
Fixed in https://github.com/FXyz/FXyz/pull/199.

PR #184, opened 2023-09-08
Migrate to extra module info plugin

samypr100:
Closes https://github.com/FXyz/FXyz/issues/180


nlisker:
Fixed in https://github.com/FXyz/FXyz/pull/199.

Issue #181, opened 2023-09-07
Use Gradle Toolchains


samypr100:
As a follow up to https://github.com/FXyz/FXyz/pull/172#discussion_r1317297881, this issue aims to capture adding support for [Gradle Toolchains](https://blog.gradle.org/java-toolchains).

nlisker:
Fixed in https://github.com/FXyz/FXyz/pull/199.

PR #185, opened 2023-09-09
Migrate to using gradle toolchains

samypr100:
Closes https://github.com/FXyz/FXyz/issues/181

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,8 +24,7 @@ subprojects {
     apply plugin: 'com.github.hierynomus.license'
 
     java {
-        sourceCompatibility = JavaVersion.VERSION_11
-        targetCompatibility = JavaVersion.VERSION_11
+        toolchain.languageVersion = JavaLanguageVersion.of(11)
         modularity.inferModulePath.set(false)
     }
 
@@ -34,6 +33,9 @@ subprojects {
         apply plugin: 'maven-publish'
         apply from: "$rootDir/gradle/publishing.gradle"
     } else {
+        java {
+            toolchain.languageVersion = JavaLanguageVersion.of(17)
+        }
         javadoc {
             enabled(false)
         }


build.gradle
```diff
@@ -34,6 +33,9 @@ subprojects {
         apply plugin: 'maven-publish'
         apply from: "$rootDir/gradle/publishing.gradle"
     } else {
+        java {
+            toolchain.languageVersion = JavaLanguageVersion.of(17)
```
jperedadnr, line 37: I take that needing JDK17 comes from the use of jfxtras-17? I'm okay having this condition here.

In any case, I see that even if I set my `JAVA_HOME` to 11, when running the samples, gradle manages to find a 17 SDK from my system and uses it, overriding my settings, which I'm not sure I like...

Also, if I set the toolchain with an SDK I don't have, for instance 18, I get `No locally installed toolchains match and toolchain download repositories have not been configured`. I see there is a plugin that can be added that will download "one", and install it under ~/.gradle/jdks, but again, I don't like this being done without user's control.

nlisker:
I'm getting the following error upon refreshing gradle:

```
FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\Nir\git\FXyz\FXyz-Samples\build.gradle' line: 58

* What went wrong:
A problem occurred evaluating project ':FXyz-Samples'.
> Could not resolve all dependencies for configuration ':FXyz-Samples:runtimeClasspath'.
   > Failed to calculate the value of task ':FXyz-Client:compileJava' property 'javaCompiler'.
      > No matching toolchains found for requested specification: {languageVersion=11, vendor=any, implementation=vendor-specific} for WINDOWS on x86_64.
         > No locally installed toolchains match and toolchain auto-provisioning is not enabled.
```

Is it only me?

nlisker:
Looks like since Gradle 8 there's a new configuration needed: https://docs.gradle.org/8.0.2/userguide/toolchains.html#sub:download_repositories

jperedadnr:
See my [comment](https://github.com/FXyz/FXyz/pull/185#discussion_r1320548326), don't you have a JDK11 installed?
I'd rather don't set a plugin to download an SDK without the user's intervention?

nlisker:
No, I migrated everything to 17+. Do we want to rely on the user having JDK11? As I see it, it's another dependency to download like the rest of them. I think that Gradle will use a locally installed one and only download one if it can't be found.

jperedadnr:
If you don't have JDK11, as it was before, with source/target = 11, the jars are built with that language level using your current JDK (17), and that is perfectly valid.

@samypr100 if I take it right, with the toolchain, this changes, and forces the users (those that build from source this library that is) to install 11 (or download it for them if we add another plugin)?

If this is the case, I don't see any benefit on adding the toolchain for the core/importers/client projects.

samypr100:
@jperedadnr Quoting from the [blog](https://blog.gradle.org/java-toolchains)

> In a nutshell, it allows you to run Gradle with whatever version of Java you have installed, while Gradle builds the project with the version of Java declared in a single place. Gradle will automatically configure the build, detect already installed JDKs and download the required JDK if it is not already installed.

There's good reasons for their decision on avoiding using, e.g. lets say JDK 17, and setting up source/target compatibility to be 11 or using JAVA_HOME only on that note. Below are the three key points of the blog on this.

> It was problematic for Gradle to build code on not yet released Java versions – like Java 16 at the time of writing – which may fail to run existing releases of Gradle – such as the current 6.7 release.

> Additionally, the code may behave differently when built and run with different Java versions, potentially causing hard to diagnose issues.

> Even without incompatibilities affecting the system behavior, using different Java versions to build the project on different machines can reduce build performance because of [build cache](https://docs.gradle.org/6.7/userguide/build_cache.html) misses.


I agreed, I do not like tooling downloading a JDK on by behalf either unless I want it to but is the direction gradle will go with in the future and eventually deprecate source/target compatibility related flags, hence the PR. We can stay course as we were before though :)

I've been using SDKMAN and toolchains picks the right version from there, although I've seen it work somtimes with system level installations (at least they claim they do support it to some extent) such as /Library/Java/JavaVirtualMachines.

I haven't had the need to configure a plugin for auto-provisioning settings and it has actually downloaded JDK's for me without configuring a plugin for that, although I do believe auto-provisioning can be flaky. If we want to block auto-provisioning explicitly, we can set `org.gradle.java.installations.auto-download=false` in the gradle.properties file.

build.gradle
samypr100: > I take that needing JDK17 comes from the use of jfxtras-17? I'm okay having this condition here.

Yes

samypr100: > Also, if I set the toolchain with an SDK I don't have, for instance 18, I get No locally installed toolchains match and toolchain download repositories have not been configured. 

I believe toolchains will only download LTS/GA releases it can find a download for. My guess is that for 18 it can't find where to download it base on your system that satisfies those constraints.

nlisker:
Fixed in https://github.com/FXyz/FXyz/pull/199.

Issue #195, opened 2025-03-09
Skybox is inside-out


rlouedec:
Skybox is inside-out: the faces seems wrongly rotated, the faces transitions are visible (sides are not matching).
Reproduced on:
- the FXSampler (Skyboxing sample or enable SkyBox in Scene Lighting of any TexturedMesh sample)
- cubemap generated by https://jaxry.github.io/panorama-to-cubemap/

![Image](https://github.com/user-attachments/assets/8d5dbb40-e63f-4c6f-b2fe-48e1654412d1)

PR #196, opened 2025-03-09
Fix Skybox faces rotations

rlouedec:
Fixes: https://github.com/FXyz/FXyz/issues/195

On FXSampler (Skyboxing sample), I noticed some glitches (gray triangles) though:
image


rlouedec:
On FXSampler, I noticed some glitches (gray triangles) though.
But only for the Skyboxing sample.
image

PR #203, opened 2025-07-06
Use gradle nexus publish plugin

### LINKED ISSUE #202: Publishing plugins discussion
As a followup to https://github.com/FXyz/FXyz/pull/199, we need to update the publishing code.

Current plugins:
https://github.com/marcphilipp/nexus-publish-plugin
https://github.com/Codearte/gradle-nexus-staging-plugin

Possible replacements:
https://github.com/gradle-nexus/publish-plugin (see https://github.com/FXyz/FXyz/issues/177)
https://github.com/jreleaser/jreleaser
Other?

Relevant links:
OSSRH EOL: https://central.sonatype.org/news/20250326_ossrh_sunset/
Reddit: https://www.reddit.com/r/java/comments/1lo3hg8/reminder_ossrh_service_endoflife_is_today/

jperedadnr:
Fixes #202 

The plugin has to be on the root project and the `nexusPublishing` block has to be added just once (keeping it inside fyxz3d-publishing seems to repeat it for the three subprojects that need it, and cause an exception).


build.gradle
```diff
@@ -1,21 +1,28 @@
...
 
 wrapper {
     gradleVersion = gradleWrapperVersion
     distributionType = Wrapper.DistributionType.BIN
 }
+
+nexusPublishing {
+    repositories {
+        sonatype {
+            nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
```
nlisker, line 20: Not sure staying on OSSRH is the right direction. Might be a good time to switch to the new API.

jperedadnr: That is the new API, isn't it?
https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central
or
https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central that mentions: 

> Replace your existing OSSRH endpoint (https://s01.oss.sonatype.org or https://oss.sonatype.org) with https://ossrh-staging-api.central.sonatype.com

nlisker: I thought https://central.sonatype.org/publish/publish-portal-gradle/ was the new API.

https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/ looks like a step from the deprecated OSSRH to the portal API:

> The Portal OSSRH Staging API is a partial reimplementation of the OSSRH / Nexus Repository Manager 2 Staging APIs. The intent of this service is to enable publishers who are using existing plugins to have a smooth migration to the Central Publisher Portal. We hope that over time plugins will adopt the [Portal API](https://central.sonatype.org/publish/publish-portal-api/) rather than rely on this service.

jperedadnr: There are two URLs that we use, and the ones that Sonatype indicates:

```
nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
```

The URL you mentioned `https://central.sonatype.org/publish/publish-portal-gradle/` is a documentation URL.

Sonatype docs mention which URLs are deprecated and which is the new one to use:

> Replace your existing OSSRH endpoint (https://s01.oss.sonatype.org or https://oss.sonatype.org) with https://ossrh-staging-api.central.sonatype.com.

So we are following what they ask...

nlisker: It's not what I understood for the documentation I linked, but if it works then it's fine.

From the [documentation](https://central.sonatype.org/publish/publish-portal-api/#authentication-authorization), I understood that the API is https://central.sonatype.com/api/v1/publisher/, which is also used in [this](https://github.com/cowwoc/anchor4j/blob/main/.github/workflows/deploy_to_maven_central.yml) project and [this](https://github.com/jskov/action-maven-publish/blob/main/src/main/java/dk/mada/action/portal/PortalProxy.java) one for example.

This [documentation](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/) says
> The Portal OSSRH Staging API is a partial reimplementation of the OSSRH / Nexus Repository Manager 2 Staging APIs. The intent of this service is to enable publishers who are using existing plugins to have a smooth migration to the Central Publisher Portal. **We hope that over time plugins will adopt the [Portal API](https://central.sonatype.org/publish/publish-portal-api/) rather than rely on this service**.

So they recommend not relying on the OSSRH staging API and using the portal API. At least this is what I understood.

Edit: Someone on reddit [confirms](https://www.reddit.com/r/java/comments/1lo3hg8/reminder_ossrh_service_endoflife_is_today/n0mjgvd/).

jperedadnr: Right, according to https://central.sonatype.org/pages/ossrh-eol/#publishing-to-central-portal, 

> the fastest way to continue publishing is the [OSSRH Staging API Service](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/) (a compatibility service that translates a subset of the Nexus 2 APIs to the Portal API)

so what I'm suggesting in this PR (which we have done already in others like the GluonFX plugins), is a temporary change after all...

and 
> If you prefer to jump right to publishing via the Portal directly, we support publishing via a [Maven plugin](https://central.sonatype.org/publish/publish-portal-maven/), [uploading a bundle](https://central.sonatype.org/publish/publish-portal-upload/), or [calling the API directly](https://central.sonatype.org/publish/publish-portal-api/). Additionally, there are a number of [third-party Gradle plugins](https://central.sonatype.org/publish/publish-portal-gradle/) that we do not support ourselves, but that you might consider.

So going directly to the Portal API we should probably use JReleaser... 
Scene Builder uses it (https://github.com/gluonhq/scenebuilder/blob/master/pom.xml#L358), but it seems quite more complex to set up that a regular gradle plugin...

Maybe you want to give it a try? Else, we can have this PR in first, and wait for someone that wants to do it later on.

nlisker: OK, so this is "quick fix". I'm fine with deferring the recommended solution to a later time.

I'll probably mess around with the new publishing mechanism at some point for my own projects and share my results here. This is how I did the build update previously.

nlisker (approved):
Didn't test the publishing, but the change looks fine. One thing I wonder about is whether all the projects get published or only the libraries, seeing as the configuration doesn't apply only to those.

jperedadnr:
You can run `sh gradlew --dry-run publish` to see that the three libraries would get published, but not the Samples, since it only includes org.fxyz3d-base, but doesn't include org.fxyz3d-lib (which has maven-publish).

nlisker:
Ah, I see.
It seems that we don't need to declare the `maven-publish` plugin in both our `lib` plugin and the `publishing` plugin since the former applies the latter. My `--dry-run publish` tests indicate that either place is fine, but it probably belongs more in the `publishing` plugin, so I suggest removing it from `lib` (line 4).

nlisker (approved):

